PR #2, opened 2022-11-28
Fix AWS not getting a pbench disk

dvalinrh:


ansible_roles/roles/aws_create/tasks/main.yml
```diff
@@ -172,17 +172,8 @@
...
+    pbench_ns_tf: pbench_disks.tf
     cloud_change_to: aws_instance
-  when: config_info.spot_start_price == 0
-
-- name: Set up terraform vars spot
-  include_role:
-    name:  set_up_tf_vars
-  vars:
-    pbench_ns_tf: tf/pbench_disks.tf
-    cloud_change_to: later
```
frival, line 184: I assume later and aws_instance are identical, either that or this was broken with 'later' all along?

frival (commented):
I assume later and aws_instance are identical, either that or this was broken with 'later' all along?

ansible_roles/roles/aws_create/tasks/main.yml
dvalinrh: They were the same. What was broken was the tf/ in front of pbench_disks.tf, that should not be there.
The two were present and left over from the original spot vs non spot work where we had things in separate locations.  With the movement to tf, that is no longer required, and figured as I was in there at that spot, to correct it all at once.

frival: Let's just be careful fixing multiple independent problems under the same commit.  It makes unwinding things much more difficult, particularly now that we're in public.  LGTM though.

PR #6, opened 2022-12-08
Fix cloud os

dvalinrh:


frival (approved):
Urf, I guess we were depending on the show_images to fail properly.  Better to be explicit, agreed.

frival (approved):
LGTM

PR #7, opened 2022-12-08
Catch error on tf create of plan.  The command returns true when ther…

dvalinrh:
…e is failures.

Check stdout for errors.


frival (approved):
LGTM

PR #9, opened 2022-12-09
Fix azure cloud times for start up

dvalinrh:
Wrong variable being used.


frival (approved):
Well that's embarrassing ...  LGTM

PR #13, opened 2023-01-19
Handle spot termination when a test is running.

dvalinrh:
Two options
1) We will create the vm without spot pricing
2) We simply terminate.

### DIFF
--- a/bin/kick_off.sh
+++ b/bin/kick_off.sh
@@ -133,13 +133,13 @@ do
 				if [[ $sp_check == *","* ]]; then
 					rm -rf tf
 					#
-					# Next attempt is with out spot pricing.
+					# Next attempt it without spot pricing.
 					#
-					grep -v "spot_range:" ansible_vars_main.yml | grep -v spot_range: > spot_repair
+					grep -v "spot_range:" ansible_vars_main.yml > spot_repair
 					echo "  spot_start_price: 0" >> spot_repair
 					echo "  spot_range: 0" >> spot_repair
 					mv spot_repair ansible_vars_main.yml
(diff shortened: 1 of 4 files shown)


sousinha97 (approved):

bin/kick_off.sh
```diff
@@ -119,6 +123,44 @@ do
...
+		if [ $spot_recover -eq 1 ] &&[[ ! -f "test_returned" ]]; then
+			#
+			# Check to see if we used spot, and we are to recover if the system goes away.
+			#
+			if [[ -f test_started ]]; then
+				sp_check=`grep spot_range: ansible_vars_main.yml`
+				if [[ $sp_check == *","* ]]; then
+					rm -rf tf
+					#
+					# Next attempt is with out spot pricing.
```
frival: micro-nit: attempt *it*

bin/kick_off.sh
```diff
@@ -119,6 +123,44 @@ do
...
+					#
+					grep -v "spot_range:" ansible_vars_main.yml | grep -v spot_range: > spot_repair
```
frival: I'm curious why you're "grep -v"-ing twice here, the input through the pipe should be excluding spot_range: already.

bin/kick_off.sh
```diff
@@ -119,6 +123,44 @@ do
...
+					# Not spot, test started, system died.
+					#
+					echo Error: test started, system died.
+					exit
+				fi
+			else
+				#
+				# Rarely will happen.
+				#
+				echo Error: did not start the test.
```
frival: We should log something to stable storage, this echo will get lost in the massive scrollback buffer any significant run creates.

frival (changes requested):
Couple issues, both "relatively" minor.

PR #14, opened 2023-01-24
Fix gp3 so it picks up on iops and througput settings.

dvalinrh:


sousinha97 (approved):

frival (approved):
I'd *prefer* to template from a single file rather than two separate ones (e.g. have a template for iops and template that results in nothing for gp2 but fills in for gp3), but that I think is just stylistic on my part.  LGTM.

dvalinrh:
had to do a hand merge, easier to do with brand new tree.

PR #15, opened 2023-01-25
Adding ability to terminate cloud instances created via terraform

dvalinrh:


sousinha97 (approved):

dvalinrh:
Fix it so Zathras can delete any cloud instance it creates.  Right now we have the option of delete after the test runs, but only then.  This adds the ability to delete the cloud instances we create and do not terminate at the end of the run (or simply create).  It requires that the --no_clean_up option is used so the tf directory does not get removed at the end of the Zathras execution.

dvalinrh:
Adding ability to terminate clouds that zathras creates but does not destroy when it is completed.

PR #16, opened 2023-01-30
Add uploads option.  When no uploads are desginated, we will use the …

dvalinrh:
…filesystem that has the most free space.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -86,7 +86,7 @@ export ANSIBLE_ROLES_PATH=$HOME/.ansible/collections/ansible_collections/pbench/
 #   gl_cloud_region_zone: zone to create the instance in.
 #
 value_not_set="none"
-gl_upload_directory="none"
+gl_kit_upload_directory="none"
 gl_host_uses_nvme=0
 gl_cloud_networks=0
 gl_cloud_net_type="default"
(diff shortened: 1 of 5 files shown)


bin/burden
```diff
@@ -1444,6 +1445,7 @@ create_ansible_options()
 		echo "  update_os_version: ${gl_update_target}" >> ansible_vars_main.yml
 		echo "  update_type: ${gl_image_type}" >> ansible_vars_main.yml
 		echo "  persistent_log: ${gl_persistent_log}" >> ansible_vars_main.yml
+		echo "  upload_directory: ${gl_upload_directory}" >> ansible_vars_main.yml
```
frival: Can we call it "kit_upload_directory" so it's more obvious that it's only for uploading kits, not e.g. other zathras bits? Or maybe "benchmark_upload_directory" or something like that?

bin/ten_of_us.yml
```diff
@@ -884,17 +905,21 @@
...
       mode: 0777
   - name: link things in
     file:
-      src: "/mnt/workloads"
-      dest: "/home/{{ config_info.test_user }}/workloads"
+      src: "{{ dyn_data.upload_dir }}/workloads"
+      dest: "/{{ config_info.user_parent_home_dir }}/{{ config_info.test_user }}/workloads"
       state: link
-    when: config_info.system_type != "local"
+#    when: config_info.system_type != "local"
```
frival: Debug stuff?

frival (changes requested):
Just a couple things.

PR #17, opened 2023-01-31
Use AWS cli to determine spot settings

dvalinrh:


bin/burden
```diff
@@ -1173,18 +1173,30 @@ aws_specific_os_version()
...
-				gl_spot_increment=`echo ${spot_string} | cut -d':' -f3`
-				gl_spot_cap=`echo ${spot_string} | cut -d':' -f4`
-			else
-				cleanup_and_exit "Error: no spot information present for $1" 1
+		spot_string=`grep $1 ${gl_test_def_dir}/spot_price.cfg`
+		if [[ $gl_system_type == "aws" ]]; then
+			#
+			# For AWS obtain the possible starting price.
+			#
+			gl_spot_price=`aws ec2 describe-spot-price-history --start-time=$(date +%s) --product-descriptions="Linux/UNIX" --query 'SpotPriceHistory[*].{az:AvailabilityZone, price:SpotPrice}' --instance-types ${1} | awk '{print $2}' | sort -n | tail -1`
```
frival, line 1181: Does Linux/UNIX match RHEL?  Since RHEL adds a slight extra fee I'm wondering how that all works out.  Maybe a non-issue?

bin/burden
```diff
@@ -1173,18 +1173,30 @@ aws_specific_os_version()
...
+				cleanup_and_exit "Error: unable to retrieve spot info for $1" 1
 			fi
+			#
+			# We will default to 5 intervals
+			# 
+			gl_spot_cap=`curl -L ec2.shop?filter=$1 | grep $1 | awk '{print $12}'`
+			gl_spot_increment=`echo "scale=2;((($gl_spot_cap - $gl_spot_price)*100)/5.00)/100" | bc`
 		else
-			cleanup_and_exit "Error: no spot information present for $1" 1
+			if [[ $? -ne 0 ]]; then
```
frival, line 1191: Do we have an issue filed to support this for Azure?  IIRC you added Azure spot, or maybe the gray hairs are multiplying faster than I thought...

frival (commented):
A couple questions, probably okay but just checking.

bin/burden
dvalinrh: Non issue, it will not be much of a difference, KISS approach.

PR #26, opened 2023-03-02
Add section to burden to handle pbench packaging.

dvalinrh:
Add packages for pbench installation.  Note simple right, now, if more packages, will convert to a list.


frival (approved):
I think I'm okay with this, but at some point (1.0 maybe, I'm not sure) pbench will not have perf in the default toolset so we should add a comment stating to review this if and when pbench removes perf from the default toolset.  No reason to install it if it's not needed right?

PR #40, opened 2023-05-30
Update tf versions for aws so newer systems with networks can reboot.

dvalinrh:
Old version is causing the i4g* and other new platforms not to reboot properly when we are dealing with internal private networks.  Bumping to the latest terraform version clears the issue out.  Not bumping azure/gcp as they are perfectly happy with the current version.


kdvalin (approved):
LGTM

PR #43, opened 2023-06-21
Custom Pbench Tool Registration

kdvalin:
This PR adds the ability to define what pbench tools are registered on remote systems.

This also supports user-tools, which can be copied over from the local system to remote systems or cloned from a git repository.

Example YAML used for configuration:
```yaml
---
interval: 10 # seconds
tools:
  - tool: iostat
  - tool: mpstat
  - tool: perf
  - tool: pidstat
  - tool: proc-interrupts
  - tool: proc-vmstat
  - tool: sar
  - tool: user-tool
    name: example
    git_url: <GIT URL> # Alternatively, replace with "path" and a path to the directory that will be copied over 
    scripts:
      start: <PATH TO START SCRIPT>  
      stop: <PATH TO STOP SCRIPT>
tool_opts:
   TOOLNAME: <TOOL OPTS>
```


dvalinrh:
General comments
1) This should be an option available via the command line
2) Need to update the user doc in documentation.

kdvalin:
Closing this out since it's way out of date.

PR #49, opened 2023-08-23
Adding azure spot.

dvalinrh:
Adding support for Azure spot.  Really wished terraform handled things better with the variables.  If priority is not spot, can not have eviction_policy in the terraform.

Also do not use network systems and spot, there is a bug where the delete will hang.  The resource deletes complete, but terraform delete never returns.


ansible_roles/roles/azure_create/files/tf/vm_spot_set
```diff
@@ -0,0 +1,28 @@
+# And finally, create the virtual machine
```
kdvalin: add `.tf` to end of file name

ansible_roles/roles/azure_create/tasks/main.yml
```diff
@@ -11,6 +11,12 @@
     copy:
       src: "tf/main_no_net.tf"
       dest: "{{ working_dir }}/tf/main.tf"
+  - name: cat non spot machine terraform
+    shell: "cat {{ config_info.local_run_dir }}/ansible_roles/roles/azure_create/files/tf/vm_spot_set | sed \"s/PRIORITYSPOT//g\" | sed \"s/EVICTIONPOLICY//g\" >> {{ working_dir }}/tf/main.tf" 
+    when: config_info.spot_start_price == 0
+  - name: cat non spot machine terraform
+    shell: "cat {{ config_info.local_run_dir }}/ansible_roles/roles/azure_create/files/tf/vm_spot_set | sed \"s/PRIORITYSPOT/priority              = \\\"Spot\\\"/g\" | sed \"s/EVICTIONPOLICY/eviction_policy       = \\\"Deallocate\\\"/g\" >> {{ working_dir }}/tf/main.tf"
```
kdvalin: Is there a reason why you're not using the `template` role within ansible?

kdvalin (changes requested):

ansible_roles/roles/azure_create/tasks/main.yml
dvalinrh: Yeah, because have run into numerous instances of it not working, or it works this month and an upgrade happens and they broke it.  So got tired of it, and simply do it this way, works all the time.  If ansible ever stabilizes enough where they do not break things that were working, will move this stuff to use the template roles.

kdvalin (approved):
LGTM

PR #53, opened 2023-08-30
Azure sporadically will not handle deletion from terraform.  Add time…

dvalinrh:
…out to the

terraform delete command for 20 minutes.


kdvalin (approved):
LGTM

PR #54, opened 2023-10-18
Add --upload_rpms option.  Need this so we can use pbench on betas in…

dvalinrh:
… the cloud

### DIFF
--- a/ansible_roles/roles/install_pbench/tasks/main.yml
+++ b/ansible_roles/roles/install_pbench/tasks/main.yml
@@ -24,6 +24,12 @@
       delegate_to: "{{ target_sys }}"
       become: yes
 
+- name: Upload and stall any rpms.
+  include_role:
+    name: install_rpms
+  vars:
+    results_file: "/tmp/rpm_installs"
+
 - name: Install pbench agent
   include_role:
     name: pbench_agent_install
--- a/ansible_roles/roles/install_pbench/tasks/main.yml
+++ b/ansible_roles/roles/install_pbench/tasks/main.yml
@@ -24,7 +24,7 @@
       delegate_to: "{{ target_sys }}"
       become: yes
 
-- name: Upload and stall any rpms.
+- name: Upload and install any rpms.
   include_role:
     name: install_rpms
   vars:
--- a/bin/burden
+++ b/bin/burden
@@ -2406,8 +2406,8 @@ convert_scenario_file()
 	#
 	# Check to make sure we have the proper number of host configs
 	# 
-	sc_cnt=`grep host_config $1 | wc -l`
-	ps_cnt=`grep host_config parse_file | wc -l`
+	sc_cnt=`grep host_config $1 | grep -o , | wc -l`
+	ps_cnt=`grep host_config parse_file | grep -o , | wc -l`
 	if [ $ps_cnt != $sc_cnt ]; then
 		cleanup_and_exit "yq did not find the appropriate number of hosts, look for duplicate names" 1
 	fi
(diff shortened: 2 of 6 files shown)


ansible_roles/roles/install_pbench/tasks/main.yml
```diff
@@ -24,6 +24,12 @@
       delegate_to: "{{ target_sys }}"
       become: yes
 
+- name: Upload and stall any rpms.
```
kdvalin:
```suggestion
- name: Upload and install any rpms.
```

bin/ten_of_us.yml
```diff
@@ -424,6 +437,21 @@
...
 
+- hosts: install_group
+  user: config_info.test_user
+  become_user: root
+  become: true
+  vars_files:
+    - ansible_vars.yml
+  gather_facts: yes
+  tasks:
+  - name: Upload and stall any rpms.
```
kdvalin:
```suggestion
  - name: Upload and install any rpms.
```

kdvalin (changes requested):
Minor typo fixes needed

dvalinrh:
Pushed set of changes
1) Fixes it so we install the rpms on all systems when network is designated
2) Fixes issue where --insall_rpms was causing us to bail out of the scenario early even though there was more to do.
3) option to use spot now so it can be used within the scenario file as it should be.

kdvalin (approved):
LGTM

kdvalin (changes requested):
GitHub did a weird thing, previous feedback still stands.

Minor typos

bin/burden
```diff
@@ -2389,6 +2403,14 @@ convert_scenario_file()
 	if [ $? -eq 1 ]; then
 		cleanup_and_exit "The file, parse_file does not meet yaml requirements." 1
 	fi
+	#
```
frival: Fair warning: I'm going to start NAKing changes that include unrelated fixes because it makes finding and reverting breakages difficult.

bin/burden
```diff
@@ -4063,7 +4096,7 @@ for i in `ls -d *`; do
 			date=`date`
 			echo "$data: $date tests completed: ${tests}, tests left ${tests_left}" >> /home/zathras_log/spot_fails
 		fi
-		sed "s/--use_spot/--no_spot_recover/g" $gl_top_dir/$file_to_use | sed "s/${tests}//g" > $gl_top_dir/rerun
+		sed "s/--use_spot 1/--no_spot_recover/g" $gl_top_dir/$file_to_use | sed "s/${tests}//g" > $gl_top_dir/rerun
```
frival: It looks like you're making "--use_spot 1" as an alias for "--no_spot_recover".  Why not just use "--no_spot_recover" where it's needed or if you have to make "--use_spot" take args at least make them symbolic rather than magic numbers?

bin/ten_of_us.yml
```diff
@@ -319,6 +319,19 @@
       exit_msg: "Code repo installation failure"
     when: (config_info.os_vendor == "rhel") and (config_info.system_type == "aws" or config_info.system_type == "azure" or config_info.system_type == "gcp")
 
+ #
+ # Determine where to upload from, assumption is all test systems have the same
```
frival, line 323: Ouch that's a heck of an assumption.

frival (changes requested):
One is a future warning, another I'd like either an explanation for or a change to.  Fair 'nuff?

bin/ten_of_us.yml
kdvalin: This one's still here :)

PR #55, opened 2023-10-26
Fix max systems

dvalinrh:
Add some sanity checking for the scenario file
Improve the speed of getting tag versions


test_system_usage
```diff
@@ -0,0 +1,3 @@
+      User                                          Run label   Instance                     Date  Price         Test   Time     Cost
+    dvalin                         arm_chip_compare_rhel_rhel m6g.xlarge      2023.10.26-11.35.34  0.154     coremark    239    .0102
+    dvalin                         arm_chip_compare_rhel_rhel m6g.xlarge      2023.10.26-12.02.12  0.154     coremark    240    .0102
```
kdvalin: This probably shouldn't be committed up.

Add it to `.gitignore` (or create one if it doesn't exist)

kdvalin (changes requested):

test_system_usage
dvalinrh: Does not belong, removing.

PR #60, opened 2023-11-27
Count actual sytems, not host config entries.

dvalinrh:
Right now we verify on host config count for verifying number systems.  We need to change this to actually check number of systems so things still work properly when


kdvalin (approved):
LGTM

PR #63, opened 2024-01-10
Fix networking for azure.  Terraform is failing on the creation of th…

dvalinrh:
…e system

### DIFF
--- a/ansible_roles/roles/azure_create/files/tf/pbench_disks.tf
+++ b/ansible_roles/roles/azure_create/files/tf/pbench_disks.tf
@@ -11,7 +11,7 @@ resource "azurerm_managed_disk" "pbench_volume" {
 resource "azurerm_virtual_machine_data_disk_attachment" "pbenchdisk" {
   count              = var.vm_count
   managed_disk_id    = element(azurerm_managed_disk.pbench_volume.*.id, count.index)
-  virtual_machine_id = element(azurerm_linux_virtual_machine.virtualmachine.*.id, count.index)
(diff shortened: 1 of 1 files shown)


ansible_roles/roles/azure_create/files/tf/pbench_disks.tf
```diff
@@ -11,7 +11,7 @@ resource "azurerm_managed_disk" "pbench_volume" {
 resource "azurerm_virtual_machine_data_disk_attachment" "pbenchdisk" {
   count              = var.vm_count
   managed_disk_id    = element(azurerm_managed_disk.pbench_volume.*.id, count.index)
-  virtual_machine_id = element(azurerm_linux_virtual_machine.virtualmachine.*.id, count.index)
+  virtual_machine_id = azurerm_linux_virtual_machine.virtualmachine[count.index].id
```
kdvalin, line 14: Is there a reason why `virtual_machine_id` needs this changed, but `managed_disk_id` doesn't?

kdvalin (changes requested):

ansible_roles/roles/azure_create/files/tf/pbench_disks.tf
dvalinrh: Different  structs.  For some reason azurerm_linux_virtual_machine use to work then stopped working.  Switching that line to not use element cleared out the variable not being seen.  The line above works fine. Welcome to the joys of ansible and terraform, where stuff works fine, then it simply stops to work.

kdvalin (approved):
LGTM

PR #65, opened 2024-02-02
Updates to have AWS use terraform for networking instead of a script.

dvalinrh:
Update to have AWS use terraform for networking instead of bash script.


ansible_roles/roles/aws_create/files/tf/networking.tf
```diff
@@ -0,0 +1,137 @@
...
+}
+
+resource "aws_vpc" "zathras_vpc" {
+    provider = aws.zathras_prov
+    enable_dns_support = true
+    enable_dns_hostnames = true
+#
+# Uncoment when using ipv6
+#
+#    assign_generated_ipv6_cidr_block = true
```
kdvalin, line 15: I don't like this, but I also cannot find a better mechanism within terraform to do it.. :/

kdvalin (approved):
Besides Terraform quibbles, LGTM

PR #67, opened 2024-02-26
Adding code to handle rerunning failed tests.

dvalinrh:


bin/burden
```diff
@@ -2503,18 +2546,13 @@ convert_scenario_file()
 
 	scenario_cnvt_file=$(mktemp /tmp/zath_convert.XXXXXX)
 	sed "s/;/----/g" $1 > $scenario_cnvt_file
+	sed_string="cat $scenario_cnvt_file"
```
kdvalin, line 2549: Repeat of line 2551

kdvalin (changes requested):

bin/burden
dvalinrh: Good catch.  Fixed.

kdvalin (approved):
LGTM

PR #69, opened 2024-03-12
Add initial install script

grdumas:
This script installs dependencies required to run Zathras.


install.sh
```diff
@@ -0,0 +1,50 @@
+#
+# Copyright (C) 2024 Greg Dumas gdumas@redhat.com
```
dvalinrh, line 4: This needs the full licensing, see front of bin/burden for it.

install.sh
```diff
@@ -0,0 +1,50 @@
...
+
+
+# pip install requirements
+pip3 install -U boto --user
+pip3 install yq --user
+
+# add yq install to path
+export PATH=/usr/local/bin:$PATH
+
+
```
dvalinrh: echo out this stuff to the screen.

install.sh
grdumas: Full license text added.

install.sh
grdumas: Which stuff should be echoed - each command being run (e.g. `dnf install ansible ...`, `pip3 install ...`, etc.)?

dvalinrh: Lines 62 to 65.
Line 64 : Should indicate only for local systems.
Line 65:  Only for existing systems (if creating clouds, do not do)
Add line:  Populate the config directory with the test templates.

grdumas: Question: why only ssh-copy-id when using an existing system? In a cloud setting, wouldn't Zathras still need to copy files to/from the system running the test?

dvalinrh: No.  As part of the cloud initialization, that is created/handled for you.

grdumas: Understood, thanks for the explanation.

grdumas: Changes implemented in script.

install.sh
```diff
@@ -0,0 +1,64 @@
...
+pip3 install yq --user
+
+# add yq install to path
+export PATH=/usr/local/bin:$PATH
+
+echo "Before you can run Zathras:
+echo "****Set up a scenario file"
+echo "****If running a local system, create the local config file for the system under test (SUT)"
+echo "****If using an existing system, ssh-copy-id between the system Zathras is installed on and the SUT."
+echo "****If working in a cloud environment do not perform the ssh-copy-id."
```
dvalinrh, line 64: Need a blurb on setting the test config templates up.  All the stuff that we git clone in into config directory (not local_config)

grdumas: Added echo to console.

install.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "****Replace the test configuration files in /zathras/config/! The default files are placeholders and do nothing!"
+echo "    If you are RH, use the internal-install script now."
```
dvalinrh: Something like this.  
Update/create the test configuration files in the config directory to reflect your requirements.

dvalinrh:
LGTM

PR #76, opened 2024-04-01
We only want to verify the last results we have.  

dvalinrh:
Multiple results in the same machine is causing false failure.


tools_bin/report_missing_failed_test
```diff
@@ -46,6 +45,7 @@ if [ -f ${curdir}/failed_test ]; then
 	do
 		while IFS= read -r expt_test
 		do
+
```
kdvalin: Did you mean to add an extra whitespace in here?

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #77, opened 2024-04-01
Need to tag AWSnetwork infra properly.  

dvalinrh:
Also need to provide a tool to delete
AWS infrastructure from a command.


kdvalin (approved):
LGTM

PR #80, opened 2024-04-10
Add check if script is run as root

grdumas:
Adds a check if the script is run as root or with sudo privileges. If the script is run with elevated privileges, it will prompt the user to confirm their choice before continuing.


bin/install.sh
```diff
@@ -17,47 +17,72 @@
...
+        [Nn]* )
+            echo "You answered no, exiting"; exit 1 ;; 
+        *) 
+            echo "Unknown input, exiting"; exit 1 ;;
+    esac
+fi
 
-# Get the operating system distribution
-os_release=$(grep "^ID=" /etc/os-release | awk -F'=' '{print $2}')
+if (( $EUID != 0 )); then
```
dvalinrh: change this to an else.  If EUID check for 0 is above, so else is automatically everything else.

dvalinrh (changes requested):
Think just that one change, when you make it I will try and see what happens.,

dvalinrh:
chmod to be 755 on the script.
Add this to output message
Make sure ~/.local/bin is in your path.
Need to also install boto3 as regular user.

PR #82, opened 2024-05-23
Fix selinux support.

dvalinrh:
Fix selinux so for Amazon and RHEL 2 ways to disable

First way is via the selinux config file (what we have been doing).
Second way is via grubby, to totally disable selinux.

Also add the ability to designated, enforcing,permissive and disabled.

For Ubuntu add code to install selinux on the OS.  Note at this time, enforcing is not supported.

### DIFF
--- a/ansible_roles/roles/selinux_config/tasks/main.yml
+++ b/ansible_roles/roles/selinux_config/tasks/main.yml
@@ -1,0 +1,94 @@
+---
+#
+# Ubuntu we have to disable and uninstall apparmor and install slinux.
+# This works on 5/17/2024
+#
+- name: Ubuntu enable
+  block:
+  - name: stop apparmor
+    shell: "systemctl stop apparmor"
+  - name: disable apparmor
+    shell:  "systemctl disable apparmor"
+  - name: uninstall app armor
+    shell: "apt-get remove apparmor -y"
+  - name: reboot the system now for disabled aparmor to take effect
+    reboot:
+      reboot_timeout: 1800
+  - name: install selinux on ubuntu
+    shell: "apt-get install policycoreutils selinux-utils selinux-basics -y"
+  - name: Activate selinux
+    shell: "selinux-activate"
+  - name: reboot the system now for selinux to be enabled.
+    reboot:
+      reboot_timeout: 1800
+  - name: We want permissive for selinux
+    shell: "setenforce 0"
+    when: (config_info.selinux_level == "permissive")
+  - name: We want enforcing for selinux
+    fail:
+      msg: "We do not support enforcing on Ubuntu at this time."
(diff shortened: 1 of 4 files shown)


ansible_roles/roles/selinux_config/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+
+- name: RHEL 2023 selinux
+  block:
+  #
+  # If we do not disable via grubby, there is still some selinux code path we call.
+  #
+  - name: general enablement
+    block:
+    - name: enable selinux RHEL/amazon
+      shell: "grubby --update-kernel=ALL --remove-args=\"selinux\""
```
kdvalin: If we're enabling SELinux why are we removing it from ALL kernels?

ansible_roles/roles/selinux_config/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+        reboot_timeout: 1800
+    when:
+      (config_info.selinux_state == "enabled")
+#
+# If we do not disable via grubby, there is still some selinux code path we call.
+#
+  - name: general disablement
+    block:
+    - name: disabling selinux for RHEL/Amazon
+      shell: "grubby --update-kernel=ALL --args=\"selinux=0\""
```
kdvalin: We should be only updating the running kernel, I think it's safe to assume most of us set the kernel we're using to the default via `grubby`

kdvalin (changes requested):

ansible_roles/roles/selinux_config/tasks/main.yml
dvalinrh: Default is for it not to be present.  If we do not do this, the cloud systems will come up with it disabled via the config file, but not the kernel.  Causes extra code path to be executed.

ansible_roles/roles/selinux_config/tasks/main.yml
dvalinrh: By default selinix is 1.  If selinux is in the grubby line, it will be set to selinux=0, which disables it all the way.

kdvalin: Maybe something a bit clearer on the `name` then.  Something like "Remove selinux disable kernel arg" or something like that.

ansible_roles/roles/selinux_config/tasks/main.yml
dvalinrh: Fixed.

ansible_roles/roles/selinux_config/tasks/main.yml
dvalinrh: Comment added.

kdvalin (approved):
LGTM

PR #84, opened 2024-05-31
Fix it so tags are assigned to all our Azure resources.

dvalinrh:


ansible_roles/roles/azure_create/tasks/main.yml
```diff
@@ -90,7 +90,14 @@
...
 - name: Add tag info in.
-  shell: "cat {{ working_dir }}/add_main_vars.tf >> {{ working_dir }}/tf/main.tf"
+  shell: "cp {{ working_dir }}/add_main_vars.tf {{ working_dir }}/temp_file.tf"
+
+- name: Add main info in.
+  shell: "cat {{ working_dir }}/tf/main.tf >> {{ working_dir }}/temp_file.tf"
+
+- name: Move file
+  shell: "cp {{ working_dir }}/temp_file.tf {{ working_dir }}/tf/main.tf"
+
```
kdvalin: You could `cat {{ working_dir }}/add_main_vars.tf {{ working_dir }}/tf/main.tf` to acheive something similar.  Overwriting the same filename seems a bit hacky

kdvalin (changes requested):

ansible_roles/roles/azure_create/tasks/main.yml
dvalinrh: have changed it so we first move main.tf, then do a cat of the two files (in appropriate order) to tf/main.tf.

ansible_roles/roles/azure_create/tasks/main.yml
```diff
@@ -89,8 +89,11 @@
 - name: copy general tf files
   shell: "cat {{ working_dir }}/add_vars_tf >> {{ working_dir }}/tf//vars.tf"
 
-- name: Add tag info in.
-  shell: "cat {{ working_dir }}/add_main_vars.tf >> {{ working_dir }}/tf/main.tf"
+- name: Move main.tf off to the side
+  shell: "mv {{ working_dir }}/tf/main.tf {{ working_dir }}//main_back.tf"
```
kdvalin: Couldn't we just rename `main.tf` to something else?  Or is it needed before these steps in the playbook?

kdvalin: If it is, fix the double `/`, will start looking for those.

kdvalin (changes requested):

ansible_roles/roles/azure_create/tasks/main.yml
dvalinrh: Renaming main.tf would impact AWS and GCP cloud pieces also, thus the moving things around.
// has been corrected.

ansible_roles/roles/azure_create/tasks/main.yml
dvalinrh: Renaming main.tf would impact AWS and GCP cloud pieces also, thus the moving things around.
// has been corrected.

kdvalin (approved):
We might want to consider fixing moving/renaming files just so we can append to them, but in a separate PR.

LGTM otherwise.

PR #87, opened 2024-06-27
Fix it so we are not rebooting the system for selinux, unless we need

dvalinrh:
Fix it so we are not rebooting the system for selinux, unless we need to.

You are all reviewers on this.  Pete indicated to do this, so folks are aware of what is going on.
We will go with the first review in.

Dave


kdvalin (approved):
LGTM

PR #88, opened 2024-07-01
Installer patch - fix yq version, correct ansible package name

### LINKED ISSUE #83: install script needs work
install script needs to
 1) Install specific versions of software
 2) Bail on errors

What i had to do earlier (edited) 
pip install 'ansible==v2.9.6'
pip install 'yq==2.10.0'
pip install jinja2==3.0.3
That was just version pieces after I ran your install script.
I had to do numerous other package installs  also.  Best way to get it all, is to install a machine clean with rhel 9, and then try things (I was on a lab machine)

grdumas:
Changes Python package installation to no longer install latest version of yq because newer versions will break Zathras functionality.

Also corrects package name 'ansible' to 'ansible-core'.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -77,7 +77,7 @@ done
 
 # pip install requirements
 pip3 install boto boto3 --user
-pip3 install 'yq==2.10.0' 
+pip3 install 'yq==2.10.0' --user 
 
 
 echo "Before you can run Zathras:"
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -34,7 +34,7 @@ else
 fi
 
 # check for and install system packages
-packages=(ansible git jq python python3-pip terraform wget)
+packages=(ansible-core git jq python python3-pip terraform wget)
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -69,14 +69,15 @@ for package in "${packages[@]}"; do
...
+        echo "package $package is not installed and not available."
     fi
 
 done
 
 
 # pip install requirements
-pip3 install -U boto boto3 yq --user
+pip3 install boto boto3 --user
+pip3 install 'yq==2.10.0'
```
kdvalin: Is there a reason why `--user` was dropped for this pip install?

kdvalin (changes requested):

bin/install.sh
grdumas: No reason, good catch. Fixed in subsequent commit.

kdvalin (approved):
LGTM

PR #90, opened 2024-07-09
Need to move the system connection before we add the pbench disk.

dvalinrh:
For some reason, if it occurs afterwards, the ssh key setup fails.


frival (approved):
Weird, but okay.  LGTM.

PR #93, opened 2024-07-23
Fix test status not always being reported properly. =

dvalinrh:
Also messasges that are bougs pertaining to results.

### DIFF
--- a/tools_bin/determine_test_status
+++ b/tools_bin/determine_test_status
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-rm initial_summary
-rm all_sys
-rm test_summary
+rm initial_summary all_sys test_summary 2>&1 /dev/null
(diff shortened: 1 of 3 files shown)


tools_bin/determine_test_status
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-rm initial_summary
-rm all_sys
-rm test_summary
+rm initial_summary all_sys test_summary 2>&1 /dev/null
```
frival, line 3: I think this would be better as rm -f rather than redirecting all error to /dev/null - that will ignore things like -EEXIST but would still give us errors like -EIO which we'd want to still collect.  It's a microscopic niggle so I'm still going to LGTM but we should discuss separately I think.

frival (approved):
LGTM with one comment for later.

PR #94, opened 2024-07-30
Remove warn line from shell command used to get CRB.  That line now

dvalinrh:
creates an error in ansible, we keep proceeding as if everything is ok, even though we did not execute the requested command.  This in turns causes the CRB not to be enabled.


frival (approved):
LGTM.  Thanks Ansible for being such a shining example of API stability.

PR #95, opened 2024-08-01
We are not finding the new results file, for ls

dvalinrh:
[We are not finding the new results file, for ls /a45ceeff709a6e20af6ac33f4a8b4bcd90c7b81a) ${sys}*/results*${expt_test}_t*tar, remove the _t


kdvalin (approved):
LGTM

PR #96, opened 2024-08-02
Check to see if the file is present before attempting removal

dvalinrh:
Error messages pertaining to files not being present when trying to remove.  Check first if the file exists then remove it.

### DIFF
--- a/tools_bin/determine_test_status
+++ b/tools_bin/determine_test_status
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-rm initial_summary all_sys test_summary 2>&1 /dev/null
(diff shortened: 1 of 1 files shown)


tools_bin/determine_test_status
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-rm initial_summary all_sys test_summary 2>&1 /dev/null
```
kdvalin, line 3: You should be able to do `rm -f` instead of looping.

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #98, opened 2024-08-02
Right now we are filtering out pbench tests by accident.

dvalinrh:
The archieve script does not know if pbench was used or not (at the level we are at, that knowledge has been lost).  So when we check for test passing, we need to remove pbench_ from the test name.  IE.  pbench_coremark will become just coremark.


kdvalin (approved):
LGTM

PR #100, opened 2024-08-08
Do not report past failures, only ones from the current run.

dvalinrh:
In the old version, we will always report past failures of the tests.  We only want to report what failed in the run we just performed.

### DIFF
--- a/tools_bin/report_missing_failed_test
+++ b/tools_bin/report_missing_failed_test
@@ -2,7 +2,10 @@
 find . -name ansible_vars_main.yml -print > ans_files
 cut -d '/' -f 2 ans_files | rev | cut -d_  -f 2- | rev | sort -u > systems
 
-rm -f initial_summary all_sys test_summary;
+if [[ -f test_errors ]]; then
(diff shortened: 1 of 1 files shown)


tools_bin/report_missing_failed_test
```diff
@@ -2,7 +2,10 @@
 find . -name ansible_vars_main.yml -print > ans_files
 cut -d '/' -f 2 ans_files | rev | cut -d_  -f 2- | rev | sort -u > systems
 
-rm -f initial_summary all_sys test_summary;
+if [[ -f test_errors ]]; then
+	cat test_errors >> past_errors
+fi
+rm -rf initial_summary all_sys test_summary test_errors tar_check failed_test
```
kdvalin, line 11: Is the `-r` needed here?

kdvalin (changes requested):

kdvalin (dismissed):
LGTM

kdvalin:
Still conflicting, remerge and I'll ack it again

PR #101, opened 2024-08-27
Allow Non RHEL OS Vendors to be run via IP

kdvalin:
This PR removes an old check enforcing the RHEL OS type when a system is specified via an IP Address.


dvalinrh:
LGTM

PR #103, opened 2024-08-30
Fix report failed tests even though the latest test passed.

dvalinrh:


frival (approved):
LGTM.  Please put some small text in the PR description to explain why this was done so in the future some poor soul will understand.

dvalinrh:
The old files from the last check are left around, and since we simply append to the file, any old failures will be reported.

PR #104, opened 2024-09-03
burden break up, large scenario lines

dvalinrh:
Move the code that is used to breakup large lines in the scenario file into a utility script.

### DIFF
--- a/bin/utils/brk_up_large_scn_host_configs
+++ b/bin/utils/brk_up_large_scn_host_configs
@@ -18,7 +18,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 #
-# This script takes a Zathras scenario file that has large lines in it and breaks
+# This script takes a Zathras scenario file that has large host lines in it and breaks
 # it down to be about 1024 characters without breaking the expected Zathras format.
 #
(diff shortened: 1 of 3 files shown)


kdvalin (changes requested):
The idea behind breaking up `burden` is not just to break it out to multiple files, but to break it out to multiple files, and each file does a specific thing that the name of the file describes.

For example: one file could be for validating the scenario file and handling error conditions specifically with the scenario file, named `validate_scenario`.

Some things like `exit_out` could be stuck in a common file which we can `source`, that way we can reuse it no matter what script we're using.

dvalinrh:
Changed new file name to something more meaningful.

bin/utils/brk_up_large_scn_host_configs
```diff
@@ -0,0 +1,231 @@
...
+# it down to be about 1024 characters without breaking the expected Zathras format.
+#
+gl_scenario_to_run=""
+
+exit_out()
+{
+	echo $1
+	exit $2
+}
+
```
kdvalin, line 31: This is fine for now, but `exit_out` will probably be used across many files, so it should be in a library shell file _somewhere_ that the scripts will `source` from.

kdvalin (approved):
LGTM, left a comment for future break-apart work.

PR #106, opened 2024-09-04
Move verify_selinux after ami list check

dvalinrh:
Having verify_selinux before listing the ami's is causing ubuntu to fail in listing the amis, unless we tell zathras that selinux is disabled.  We are not bringing a system up, only listing the amis.  No reason to do that check.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -4257,8 +4257,6 @@ else
 
 fi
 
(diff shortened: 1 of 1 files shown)


frival (approved):
LGTM

PR #108, opened 2024-09-05
Add utility for Zathras to exit.

dvalinrh:
Setup a generic exit script for Zathras.

### DIFF
--- a/bin/utils/cleanup_and_exit
+++ b/bin/utils/cleanup_and_exit
@@ -1,0 +1,186 @@
+#!/bin/bash
+#                         License
+#
+# Copyright (C) 2024  David Valin dvalin@redhat.com
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


bin/utils/cleanup_and_exit
```diff
@@ -0,0 +1,186 @@
...
+# 1) Delete the lock file
+# 2) Restore a set of files.
+# 3) Remove the cli_supplied options file.
+#
+# At the end we will exit out with the return code passed.  If no return
+# code is passed, then a value of 0 is returned by default.
+#
+# Note this script should be called via exec, that way when we exit, we will exit
+# back to the shell, not the calling program.
+#
```
kdvalin: Instead of requiring an `exec` maybe we should refactor this so we `source` the exit code into the original script, and then call exit.

This way we don't need to do anything special to ensure that the script dies when it should.

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #109, opened 2024-09-06
Move the test version checking and updates to utility script

dvalinrh:
Moving the version checking and updating of the test wrappers into own script.


bin/burden
```diff
@@ -2813,95 +2813,7 @@ verify_test_configuration()
...
-				rm -rf $temp_dir 
-			else
-				echo $test_repo, Only github is supported.
-			fi
-		done
-		popd > /dev/null
-		if [ $gl_update_test_versions -eq 1 ]; then
-			cleanup_and_exit "Update of test versions completed" 0
-		fi
+		source $UTILS_DIR/test_versions_ops --update_test_versions $gl_update_test_versions --test_version_check $gl_test_version_check --config_dir config
```
kdvalin, line 2816: Is `source` needed here?

bin/utils/test_versions_ops
```diff
@@ -0,0 +1,199 @@
...
+		echo "option not found ${1}"
+			usage $0
+		;;
+	esac
+done
+
+test_version_ops
+
+if [ $update_test_versions -eq 1 ]; then
+	source ${UTILS_DIR}/cleanup_and_exit_out --msg_string "Update of test versions completed" --rtc 0
```
kdvalin, line 201: Are we intending to kill off the parent shell?  This script is `source`'d in, which will inherit the parent shell, sourcing `cleanup_and_exit_out` with rtc 0 will kill the entire shell.

bin/utils/test_versions_ops
```diff
@@ -0,0 +1,199 @@
...
+						fi
+						echo $replace_line >> ${test_repo}.new
+					done < "${test_repo}"
+					mv $test_repo ${test_repo}.orig
+					mv ${test_repo}.new ${test_repo}
+				fi
+			fi
+			rm -rf $temp_dir 
+		else
+			echo $test_repo, Only github is supported.
```
kdvalin: Should be a non-zero exit code

kdvalin (changes requested):

bin/burden
dvalinrh: Yes, we can exit out in the script.  We can move that logic back in, but it will be duplicated in both locations.

bin/utils/test_versions_ops
dvalinrh: Fixed.

bin/utils/test_versions_ops
dvalinrh: It kills the caller, not the shell this is executed from (just verified).

bin/utils/test_versions_ops ....
does not kill the shell we ran from.  We need this so burden will exit out when it is suppose to.

kdvalin (approved):
LGTM

PR #110, opened 2024-09-10
We should not be deleting the tf directory if we have designated

dvalinrh:
that the system is not to be destroyed at the end of testing.


kdvalin (approved):
LGTM

PR #111, opened 2024-09-10
Move the terraform delete and terraform list to util

dvalinrh:
Move the terraform list and terraform delete code from burden into its own script.  burden will go these scripts to perform the operation.  Note, this is only used if we told Zathras not to tear down the cloud image upon completion of the test.  Zathras will still use the ansible roles tf_delete if we are to tear down the image at the completion of the test.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -4029,14 +4029,16 @@ fi
 # Does not return
 #
 if [ $gl_tf_terminate -eq 1 ]; then
-	source $UTILS_DIR/terraform_ops --terminate_list $gl_tf_term_list
+	$UTILS_DIR/terraform_ops --terminate_list $gl_tf_term_list
+	cleanup_and_exit "" 0
 fi
 
 #
 # Does not return
 #
 if [ $gl_tf_list -eq 1 ]; then
-	source $UTILS_DIR/terraform_ops --tf_list
+	$UTILS_DIR/terraform_ops --tf_list
+	cleanup_and_exit "" 0
 fi
 
 if [[ $gl_run_prefix == "" ]]; then
--- a/bin/burden
+++ b/bin/burden
@@ -133,7 +133,7 @@ gl_preflight=0
 #   gl_top_dir: Directory we are executing from
 #
 gl_tf_terminate=0
-gl_tf_term_list=""
+gl_tf_term_list="all"
 gl_tf_list=0
 gl_sys_file=$value_not_set
 gl_cloud_execute_tests=1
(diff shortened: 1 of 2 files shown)


bin/utils/terraform_ops
```diff
@@ -0,0 +1,165 @@
...
+		echo $tf_term_list
+		echo $dirs
+	fi
+	for tf_del in $dirs; do
+		pushd $tf_del > /dev/null
+		terraform plan -var-file=env.tfvars -destroy -out=destroy.tfplan
+		if [ $? -eq 0 ]; then
+			terraform apply "destroy.tfplan"
+		else
+			echo Warning: Unable to create the plan for $tf_del
```
kdvalin: Surround message with double quotes

kdvalin (changes requested):

bin/burden
```diff
@@ -4104,14 +4029,14 @@ fi
 # Does not return
 #
 if [ $gl_tf_terminate -eq 1 ]; then
-	tf_terminate
+	source $UTILS_DIR/terraform_ops --terminate_list $gl_tf_term_list
```
kdvalin: Does this need to be `source`'d in?

bin/burden
```diff
@@ -4104,14 +4029,14 @@ fi
...
 fi
 
 #
 # Does not return
 #
 if [ $gl_tf_list -eq 1 ]; then
-	tf_list
+	source $UTILS_DIR/terraform_ops --tf_list
```
kdvalin: Is `source` needed here?

bin/utils/terraform_ops
```diff
@@ -0,0 +1,165 @@
...
+		echo $tf_term_list
+		echo $dirs
+	fi
+	for tf_del in $dirs; do
+		pushd $tf_del > /dev/null
+		terraform plan -var-file=env.tfvars -destroy -out=destroy.tfplan
+		if [ $? -eq 0 ]; then
+			terraform apply "destroy.tfplan"
+		else
+			echo "Warning: Unable to create the plan for $tf_del"
```
kdvalin: Should be exiting with non-zero code.  If this script needs to be `source`'d, it's not needed due to killing off the parent shell.

kdvalin (changes requested):

bin/burden
dvalinrh: Source is required because we want burden to terminate when completed.

bin/burden
dvalinrh: Yes we want burden to terminate when the script is done.  We could duplicate the logic, but  then would have to deal with rtcs coming back.

bin/utils/terraform_ops
dvalinrh: Needs to use cleanup_and_exit_out here and all places we are exiting with errors.

bin/burden
frival: Wouldn't it make more sense to have this logic as a function, then source the script as if it's a library and call the function?

dvalinrh: Part of the idea, is these are also standalone.  So you are able to call terraform_ops outside of burden.

kdvalin: If it was truly standalone, it would be essentially a script you call, check the exit code of, and handle appropriately.  As it is now, it relies on stuff within Zathras (`cleanup_and_exit`)

kdvalin: Let's nudge this one.

I think it would be clearer for this to be it's own standalone script that can be used elsewhere.

dvalinrh:
Have removed source, and we simply exit after the operation.

kdvalin (commented):
This LGTM, let's go over it at the next meeting and then ship it

bin/utils/terraform_ops
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
+#                         License
+#
+# Copyright (C) 2024  David Valin dvalin@redhat.com
```
dvalinrh: update date

kdvalin (approved):
LGTM

PR #113, opened 2024-09-11
Java updates

frival:
This adds support for Java 17 and Java 21.  Package names have been verified by booting the respective OSes in AWS, and I've run a test on a local bare metal host.  One thing this does not do (which the current code does not either) is _force_ the benchmark to use the selected Java version through alternatives on RHEL (I'm assuming there's a method on Ubuntu as well).  There's an odd corner case I've run into where that's problematic (when multiple versions are already installed but the alternative points to the "wrong" one).  Given it's a corner case and easily worked around manually I'd like to see that addressed in a separate PR.


dvalinrh:
LGTM

PR #117, opened 2024-09-13
Couple minor fixes.

dvalinrh:
If we are using a local system, do not create the tf directory.
Add a space after TicketID: in tags.conf


frival (approved):
LGTM

PR #118, opened 2024-09-16
Do not install packages, let the user or config files handle it.

dvalinrh:
Do not install lksctp-tools-devel for Ubuntu, package not present.
Update phoronix test config file to point to the latest.


kdvalin (approved):
LGTM

PR #119, opened 2024-09-18
Add missing speccpu fprate test

dvalinrh:
The speccpu config file is missing the fprate test, add it in.


frival (approved):
Well that's embarrassing that we all missed it in the review.

PR #120, opened 2024-09-23
Do not view nfs fs as available to use.

dvalinrh:
We do not want to consider NFS filesystems as available to upload files to.  Filter out via df --local.


frival (approved):
With confirmation it's valid on both RHEL- and Ubuntu-based systems, LGTM.

PR #123, opened 2024-10-14
Azure: Pull the publisher from the passed urn.

dvalinrh:
Azure
1) Pull the publisher from the passed urn.
2) Verify the built urn is actually valid.


frival (commented):
Could you change / remove the comment at 1254-1255?  I don't think it applies any more.  After that I think I should be able to approve.

dvalinrh:
Comments removed.

frival (approved):
LGTM

Issue #75, opened 2024-03-14
Fix zathras so it can have multiple instances running the same time within the directory  (temp files)


dvalinrh:
We want the ability to run multiple scenario files at the same time in the same directory.

PR #124, opened 2024-10-21
Multi run

dvalinrh:
Update zathras to allow for simultaneous  execution of numerous zathras scenarios from the same run directory. 

Issue #75 
Relates to JIRA: RPOPC-261


bin/burden
```diff
@@ -280,6 +286,7 @@ cleanup_and_exit()
 			rtc=1
 		fi
 	fi
+	rm -f $gl_cli_supplied_options $gl_run_info ${gl_run_info}.sorted
```
kdvalin, line 358: Love seeing the temporary files removed after we're done with them :)

bin/burden
```diff
@@ -2770,6 +2807,7 @@ run_scenario()
 	if [[ $gl_update_target != $value_not_set ]]; then
 		$UTILS_DIR/cleanup_install_lock $BASHPID
 	fi
+	rm $scenario
```
kdvalin: Are we intending to remove the scenario file here?

kdvalin (commented):
Just some minor questions

bin/burden
```diff
@@ -1977,12 +2004,12 @@ create_ansible_options()
...
 							grep -q Failed $results_report
 							if [ $? -ne 1 ]; then
-								echo "${timestamp} Error: System: ${test_sys}, Test, $i, reported failure" >> $gl_top_dir/failed_runs
+								flock -x $gl_top_dir/failed_runs echo "${timestamp} Error: System: ${test_sys}, Test, $i, reported failure" >> $gl_top_dir/failed_runs
 								echo "    Error: reported a failure" >> ../results_info
 							fi
 						fi
 						rm $tfile
-					done
+					done
```
dvalinrh: Added space here.

bin/burden
```diff
@@ -2488,28 +2519,33 @@ convert_scenario_file()
...
 	fi
 	#
 	# Verify the file just created is valid.
 	#
-	sed "s/----/;/g" parse_file.tmp > parse_file
-	python -c 'import yaml, sys; print(yaml.safe_load(sys.stdin))' < parse_file
+	sed "s/----/;/g" $parse_file_tmp > $gl_parse_file
+	python -c 'import yaml, sys; print(yaml.safe_load(sys.stdin))' < $gl_parse_file
 	if [ $? -eq 1 ]; then
+		rm -f $parse_file $gl_parse_file_tmp
```
dvalinrh: missing removal for parse pid  fileon failure

kdvalin: gl prefix should be switched `gl_parse_file` and `parse_file_tmp`

bin/burden
dvalinrh: change scenario to be wrking_scenario instead.  Less confusion later.

bin/burden
```diff
@@ -4003,9 +4039,13 @@ first_invocation()
 		fi
 	fi
 
-	integrate_templates ${gl_test_def_dir}/test_defs.yml
-	cat $gl_test_def_dir/full_test_defs.yml | yq . > test_info
-	cat $gl_test_def_dir/java_pkg_def.yml | yq . > java_info
+	flock -x 0
+	if [[ ! -f test_info ]]; then
```
dvalinrh: Only do this once,m not each time.

dvalinrh: The test_info check makes sure we do it once.

bin/burden
```diff
@@ -209,6 +211,10 @@ gl_cpu_type_request="none"
 gl_run_file=""
 gl_failed_test_rpt="failed_tests"
 
+#
+# Make sure the lock directory exists
+#
+mkdir -p ${gl_top_dir}/lock_dir
```
kdvalin: Does this `lock_dir` ever get cleaned up?  I don't see a `rmdir` in this PR

kdvalin (changes requested):

bin/burden
dvalinrh: Fixed.

bin/burden
dvalinrh: No, as we have no clue when we are the last one, also you may start another burden when that one is finishing up.

bin/burden
dvalinrh: Good catch.

kdvalin:
@dvalinrh This has a merge conflict, resolve it and I'll re-review.

dvalinrh:
Add a lock comment for when we need it and do not need it.

bin/burden
```diff
@@ -231,24 +258,71 @@ else
...
+#     this prevents that from happening if we do not own the lock.
+#
+gl_lock=${gl_top_dir}/gl_lock
+gl_have_locked=0
+gl_debug_lock=0
+
+gl_lock_exclusive()
+{
+	if [[ $gl_debug_lock -eq 1 ]]; then
+		echo lock $1
```
frival: So the arg passed is purely for debug?  It'd be good to note that, though I'm still wondering if this is going to cause an issue at some point down the road but maybe it's good enough for now?

frival (commented):
Mostly looks good, but I left a question about the number passed to the lock functions.

bin/burden
dvalinrh: Have changed it so we report the calling function and line number via
echo ${FUNCNAME[1]} ${BASH_LINENO[0]}
nothing needs to be passed into the lock functions.

bin/burden
```diff
@@ -1901,10 +1992,24 @@ create_ansible_options()
 			if [[ $gl_ansible_verbosity != "normal" ]]; then
 				base_string=${base_string}" -l $gl_ansible_verbosity"
 			fi
+			echo $run_dir >> /tmp/dave
```
kdvalin: Remove debug statement

kdvalin (changes requested):
Found a left-over debug statement, remove it and I'll approve

kdvalin (approved):
LGTM

PR #126, opened 2024-10-28
Remove pbench from zathras.

dvalinrh:
Remove pbench support.


kdvalin:
Let's wait this one out until we have PCP metrics available.  If pbench is broken and needs manual remediation, document it to let current users know how to work around it.

PR #127, opened 2024-10-29
We only want to report what failed in the run we just performed.

dvalinrh:


tools_bin/report_missing_failed_test
```diff
@@ -25,7 +31,8 @@ do
 		tar xf ../$tball $file 2> /dev/null
 		grep -q Ran $file
 		if [ $? -ne 0 ]; then
-			echo "Failed: $sys $expt_test" >> ${curdir}/failed_test
+			fail_run=`echo $tball | cut -d'/' -f 1`
+			echo "Failed: $fail_run $expt_test" >> ${curdir}/failed_test
```
kdvalin: Maybe have a timestamp for the past errors.  That way we can tell if an error has occured recently or not

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #128, opened 2024-10-30
Add journalctl -a to boot info.

dvalinrh:
Add journalctl -a output to boot timings.  Right now we are not getting the early boot time information, this will give it to us.


kdvalin (approved):
LGTM

PR #129, opened 2024-11-18
Update Zathras installer

grdumas:
The merges in this pull request 1) installs the Ansible AWS collection and 2) freezes the terraform package to version 1.9.8-1.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -70,6 +70,7 @@ for package in "${packages[@]}"; do
         sudo dnf install terraform-1.9.8-1 -y
     else
         echo "package $package is not installed and not available."
+        exit 1
     fi
 
 done
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -17,6 +17,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
+set -eu
+
 # Check if script is being run as root
 if (( $EUID == 0 )); then
     read -p "For most use cases, running this script as root is NOT recommended. Are you sure? Y/N" yesno
@@ -39,9 +41,6 @@ packages=(ansible-core git jq python python3-pip terraform wget)
 for package in "${packages[@]}"; do 
     if dnf list installed "$package" &> /dev/null; then
         echo "$package is installed."
-    elif dnf list "$package" &> /dev/null; then
-        echo "$package is not installed but available. Installing..."
-        sudo dnf install -y "$package"
     elif [ $package == "terraform" ]; then
         # Add the terraform repository from HashiCorp
         # currently supported distros: fedora, RHEL
@@ -67,10 +66,14 @@ for package in "${packages[@]}"; do
         sudo dnf config-manager --add-repo $repo_url
 
         # install the package
-        sudo dnf install terraform-1.9.8-1 -y
+        sudo dnf install terraform-1.9.8-1 -y || {
+            exit 1
+        }
     else
-        echo "package $package is not installed and not available."
-        exit 1
+        echo "Installing $package..."
+        sudo dnf install -y "$package" || {
+            exit 1
+        }
     fi
 
 done
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -21,7 +21,7 @@ set -eu
 
 # Check if script is being run as root
 if (( $EUID == 0 )); then
-    read -p "For most use cases, running this script as root is NOT recommended. Are you sure? Y/N" yesno
+    read -p "For most use cases, running this script as root is NOT recommended. Are you sure? Y/N " yesno
 
     case $yesno in
         [Yy]* )
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -80,6 +80,10 @@ pip3 install boto boto3 --user
 pip3 install 'yq==2.10.0' --user 
 
 
+# install AWS collection for ansible
+ansible-galaxy collection install amazon.aws
+
+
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -67,7 +67,7 @@ for package in "${packages[@]}"; do
         sudo dnf config-manager --add-repo $repo_url
 
         # install the package
-        sudo dnf install terraform -y
+        sudo dnf install terraform-1.9.8-1 -y
```
frival: Wouldn't some error checking be wise?  The same would go for the above `dnf install` line.

bin/install.sh
```diff
@@ -80,6 +80,10 @@ pip3 install boto boto3 --user
 pip3 install 'yq==2.10.0' --user 
 
 
+# install AWS collection for ansible
+ansible-galaxy collection install amazon.aws
```
frival: @dvalinrh do we still need this or was it only for the days of yore when we were using the ansible cloud modules?

frival (changes requested):
Suggesting adding error detection, but perhaps you want that as a separate PR - I'll leave that to you.

bin/install.sh
dvalinrh: No, we should not need that.  Recommendation is simply to comment it out for now.

bin/install.sh
kdvalin: Adding onto this, it might be wise to have the versions recorded off in a different file (so we can source it and use $TF_DNF_VERSION or something).

Makes it easier to bump versions in the future if needed

grdumas: I've added the error checking in. 

I think the log file is a good idea - in the interest of keeping the scope of this PR focused I've opened a new issue for it (#137).

grdumas:
# Description
This PR fixes #131 by adding the ansible-galaxy installation of amazon.aws to the installation process. 

Per reviewer request, error handling has been added to the script via `set -eu` and the use of exit conditions.

# Before/After Comparison
## Package installation: no error handling
```
[root@localhost ~]# ./zathras/bin/install.sh
For most use cases, running this script as root is NOT recommended. Are you sure? Y/Ny
You answered yes, continuing install as root.
package ansible-core is not installed and not available.
git is installed.
jq is installed.
package python is not installed and not available.
package python3-pip is not installed and not available.
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use "rhc" or "subscription-manager" to register.

Adding repo from: https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
Updating Subscription Management repositories.
Unable to read consumer identity
...
```

## Package installation with error handling:
```
[root@localhost ~]# ./zathras/bin/install.sh
For most use cases, running this script as root is NOT recommended. Are you sure? Y/Ny
You answered yes, continuing install as root.
Installing ansible-core...
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use "rhc" or "subscription-manager" to register.

Error: There are no enabled repositories in "/etc/yum.repos.d", "/etc/yum/repos.d", "/etc/distro.repos.d".
```

## pip3 package installation: no error handling
```
[root@localhost ~]# ./test_pip_install.sh 
ERROR: Could not find a version that satisfies the requirement dfshjksdfhsd (from versions: none)
ERROR: No matching distribution found for dfshjksdfhsd
Collecting boto
  Using cached boto-2.49.0-py2.py3-none-any.whl (1.4 MB)
Installing collected packages: boto
Successfully installed boto-2.49.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

## pip3 package installation with error handling:
```
[root@localhost ~]# ./test_pip_install.sh 
ERROR: Could not find a version that satisfies the requirement dfshjksdfhsd (from versions: none)
ERROR: No matching distribution found for dfshjksdfhsd
```

## ansible collection installation: no error handling
```
[root@localhost ~]# ./test_ansible_install.sh 
ERROR! Neither the collection requirement entry key 'name', nor 'source' point to a concrete resolvable collection artifact. Also 'name' is not an FQCN. A valid collection name must be in the format <namespace>.<collection>. Please make sure that the namespace and the collection name contain characters from [a-zA-Z0-9_] only.

Could not find jhfkadsjfask.
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Downloading https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/amazon-aws-9.1.0.tar.gz to /root/.ansible/tmp/ansible-local-36272ie1mzeel/tmpk4_vo5ev/amazon-aws-9.1.0-eri6n6f4
Installing 'amazon.aws:9.1.0' to '/root/.ansible/collections/ansible_collections/amazon/aws'
amazon.aws:9.1.0 was installed successfully
[root@localhost ~]#
```

## ansible collection installation with error handling:
```
[root@localhost ~]# ./test_ansible_install.sh 
ERROR! Neither the collection requirement entry key 'name', nor 'source' point to a concrete resolvable collection artifact. Also 'name' is not an FQCN. A valid collection name must be in the format <namespace>.<collection>. Please make sure that the namespace and the collection name contain characters from [a-zA-Z0-9_] only.

Could not find jhfkadsjfask.
```


# Clerical Stuff
Relates to JIRA: RPOPC-279

bin/install.sh
grdumas: Turns out we still needed it.

dvalinrh (approved):

frival (approved):

PR #132, opened 2024-12-09
Need to install the ansible galaxy package community.aws

dvalinrh:
We are missing the ansible galaxy aws package on the install.  This causes zathras to fail with AWS clouds, in a way that does not really point to anything.

Fix the issue by updating the install script toinstall the appropriate ansible galaxy package.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -79,7 +79,16 @@ done
 pip3 install boto boto3 --user
 pip3 install 'yq==2.10.0' --user 
 
-
+#
+# Install required galaxy packages.
+#
+ansible-galaxy collection list | grep -q amazon.aws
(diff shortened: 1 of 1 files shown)


grdumas:
A few weeks ago in #129 I put in the code to install ansible-galaxy, but @dvalinrh you replied that the package would not be needed. Did that assessment change?

If the package is needed, I can finish up the requested changes on that PR so we can merge.

dvalinrh:
My oops, we still need it.  We do not need it for the creation but need it for gathering various aspects of the configuration.

grdumas:
Understood - I will finish work on #129.

dvalinrh:
Will be handled via another pr

PR #136, opened 2024-12-11
Add issue and PR templates

kdvalin:
This PR adds an issue template and a PR template to make it easier for contributors to follow our conventions.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,14 @@
+# The problem
+Write a short summary about what the reported problem/enhancement is.
+
+# Expected Behaviour
+How should the program behave when it encounters
+this issue?
+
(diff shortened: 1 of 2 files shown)


frival (approved):
I'm mostly okay with this; I'm not sure hardcoding RPOPC into the template is the right choice.  I'd like at least one or two other opinions before pushing though.

kdvalin:
Is there a case where we wouldn't be using RPOPC as the JIRA project?

dvalinrh:
I do not think it is a big issue.  In the future we will probably use something else, but it is not something that will change at any frequent interval.  When it does change, we will have to update many of our tools anyways.

dvalinrh (approved):
Looks fine.

Issue #133, opened 2024-12-10
ansible error when packaging install fails


dvalinrh:
We are getting an error on a variable when package installs fail.  
TASK [package_install_error : indicate terminate system] ************************************************************************************************************************************************************************************************
fatal: [new_hawkeye]: FAILED! => {"msg": "The conditional check 'config_info.error_repo_errors == 1 and rtc_val != \"0\"' failed. The error was: error while evaluating conditional (config_info.error_repo_errors == 1 and rtc_val != \"0\"): {{ install_results.rc }}: 'dict object' has no attribute 'rc'. 'dict object' has no attribute 'rc'. {{ install_results.rc }}: 'dict object' has no attribute 'rc'. 'dict object' has no attribute 'rc'\n\nThe error appears to be in '/home/dvalin/quiet_ansible/ansible_roles/roles/package_install_error/tasks/main.yml': line 6, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n  - name: indicate termi

As part of this fix we will move the error termination into package_install_error.  Doing so eliminates numerous calls to other ansible roles.

dvalinrh:
Has been fixed.

PR #139, opened 2024-12-17
Fix issue with difference in ansible yum output.

dvalinrh:
Issue #133  

Problem Description 

The ansible yum module is returning different contents in the status register  when failing on a a straight package install versus "@Development Tools".  The later yum operation when returning an error does not contain the variable "rc" in the status register, and results in the playbook terminating on syntax and not cleaning up any of the outstanding clouds.

Solution

1) Remove checking the status value set via ansible yum module.
2) Use a rescue block in the role "install_packages" to handle when the yum module fails.
3) In ten_of_us
     -  download the yum install  status file, and check to see if we have failed.
     - remove duplicate install_status check
     - Always check the install status, regardless if local system or other.

Issue #133 
Relates to JIRA: RPOPC-320


ansible_roles/roles/install_packages/tasks/main.yml
```diff
@@ -7,18 +7,25 @@
...
-  - name: Error check
-    include_role:
-      name: package_install_error
-    vars:
-      rtc_val: "{{ install_results.rc }}"
-      packages: "{{ pkg_install }}"
+  - name: Report Success
+    lineinfile:
+       path: "{{ results_file }}"
+       line: "status: sucess"
```
dvalinrh: typo

kdvalin (commented):
Will approve once the typo is addressed

kdvalin (approved):
LGTM

kdvalin (approved):
LGTM

Issue #50, opened 2023-08-23
boto check should only happen if using AWS


frival:
No reason to force people to install boto if they're not using Zathras on AWS.

https://github.com/redhat-performance/zathras/blob/56acc03bd09c1acd7bcaf5a21c8fd1e1871d08ef/bin/burden#L2045

dvalinrh:
Has been fixed.

PR #143, opened 2025-01-07
Only check for boto3 if we are doing aws.

dvalinrh:
We only need boto3 for aws, do not perform the check if system type is anything else.

Issue #50
Relates to JIRA: RPOPC-269


kdvalin (approved):
LGTM

PR #144, opened 2025-01-10
Change SELinux defaults depending on os_vendor

### LINKED ISSUE #114: `selinux_level` default incompatible with Ubuntu
The `selinux_level` default is `enforcing`.  When running against a Ubuntu host, Zathras errors out saying something to the effect of `enforcing is not supported on Ubuntu`.

When no arguments to the contrary are supplied (`selinux_level` or `selinux_state=enabled`), selinux should not be installed and the default for `selinux_level` should not be `enforcing` as it is not a supported configuration.

kdvalin:
This PR changes the default SELinux state depending on the OS vendor.  This PR will only set `selinux_state` to be default `enabled` if the target OS is RHEL.  All other vendors will default to `disabled`.

This fixes #114.

Relates to JIRA: RPOPC-260

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -165,7 +165,7 @@ gl_run_prefix=""
 gl_scenario_to_run=""
 gl_scenario_to_restore=""
 gl_selinux_level="enforcing"
-gl_selinux_state="enabled"
+gl_selinux_state=$value_not_set
 gl_selinux_state_set=0
 gl_ssh_key_file=""
 gl_show_os_versions=0;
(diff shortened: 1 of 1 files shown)


bin/burden
```diff
@@ -4073,6 +4073,14 @@ if [[ $gl_show_os_versions -eq 1 ]]; then
 	cleanup_and_exit "" 0
 fi
 
+if [[ $gl_selinux_state == $value_not_set ]]; then
```
dvalinrh: Double quote strings please.

dvalinrh (changes requested):

bin/burden
```diff
@@ -4073,6 +4073,14 @@ if [[ $gl_show_os_versions -eq 1 ]]; then
 	cleanup_and_exit "" 0
 fi
 
+if [[ "$gl_selinux_state" == "$value_not_set" ]]; then
+	if [[ "$gl_os_vendor" == "rhel" ]]; then
```
dvalinrh: include amazon linux for enabled.

dvalinrh (changes requested):

dvalinrh (approved):
LGTM

Issue #141, opened 2025-01-02
Check pip3 and dnf for awscli


dvalinrh:
awscli may be provided either by pip3 or dnf, we need to check both locations.

dvalinrh:
Fixed.

PR #146, opened 2025-01-14
Fix the aws cli check to check the existance of the command.

dvalinrh:
# Description
Fixes so we check for the existence of the aws command, not if it was installed via dnf.

# Before/After Comparison
Before the change, if the aws command was installed from pip, we would fail as we were checking to see if the package had been installed via dnf.
After the change, we check to see if the aws command is actually present, if not, we flag the error.

# Clerical Stuff
Issue: #141 
Relates to JIRA: RPOPC-282

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -2092,10 +2092,11 @@ check_for_python()
 
 check_for_aws()
 {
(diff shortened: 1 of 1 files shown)


kdvalin (approved):
LGTM

PR #145, opened 2025-01-13
Add PR Requirement Checking

kdvalin:
# Description
This PR adds a github action workflow to check for properly structured PRs.  This will check for a reference to a GitHub issue within the same repository, and a ticket within the relevant board on our Jira instance.

This will trigger when a PR is opened, and when the PR description has been edited.

# Before/After Comparison
## Before
 A PR could be submitted and contain no references to any Jira ticket or GitHub issue

## After
A PR needs a Jira reference along with a reference to a GitHub issue.  Jira ticket numbers are not checked due to permission headaches.

There is no associated github issue or Jira ticket with this PR.

PR #147, opened 2025-01-15
Add Issue Labelling based on PR state

kdvalin:
# Description
This PR adds GitHub Action workflows such that other PRs will sync their state with its parent issue using a label.  This is so we can set up Sync2Jira to automatically transition Jira tickets to their relevant stages.

Note:  This workflow will **_NOT_** work with forked repositories due to limitations on the `GITHUB_TOKEN` when certain events are triggered by a forked repository.

# Before/After Comparison
Before this PR, Jira tickets would need to be manually set to their state.

After this PR, the parent issue will receive a label, and once configured, Sync2Jira will update the Jira ticket with the appropriate label, which can then trigger an Automation moving the ticket.

This has a bit of overlap with #145 since it also checks for a parent issue in a PR.


dvalinrh (changes requested):
Please add documentation .

dvalinrh (dismissed):
LGTM

dvalinrh (approved):
LGTM

Issue #148, opened 2025-01-15
Need to be able to ssh into both system with networks present.


dvalinrh:
# The problem
New uperf version will ssh into either system.  The current setup, only allows it to the remote system, not thelocal./

# Expected Behaviour
How should the program behave when it encounters
this issue?
It should be able to ssh into either system.

# Actual Behaviour
You can not log in from the system a as root back into system a via ssh

dvalinrh:
Fixed.

PR #149, opened 2025-01-15
Uperf requires us to be able to log back in to ourselves.  Fix ssh se…

dvalinrh:
…tup.

# Description
Fixes it so when we set up the networks in the cloud we may ssh back into the system we are logged into.

# Before/After Comparison
Before change:  Uperf test will fail because we can not ssh into the system we are ssh'ing from.
After change: Uperf test works fine.,

# Clerical Stuff
Issue: #148 

Relates to JIRA: RPOPC-293


kdvalin (approved):
LGTM

PR #153, opened 2025-01-17
Add label checking for group reviews

kdvalin:
# Description
This PR adds a workflow to prevent merging without a label being applied indicating the PR has gone through a group code review.

# Before/After Comparison
Before this PR, there is nothing preventing anyone from merging a PR without going through a group code review. 

After this PR, the PR will need an additional label (currently `group_review_lgtm`) before a merge can happen.  This will allow others to submit approving code reviews before a group code review meeting and still prevent merging until the label has been applied to the PR.

There is no Issue number or Jira ticket associated with this PR.


frival (dismissed):
LGTM.  We should provide a gh CLI-based tool or alias example to add the label - it's as simple as `gh pr edit --add-label <pr>` I believe,

frival (approved):
LGTM, and thanks for adding the CLI example - let's make sure @dvalinrh knows about that for the review meetings he runs.

Issue #154, opened 2025-01-21
Azure instance creations are failing


dvalinrh:
Azure has changed again what terraform has to pass to create an instance.

We now need to pass our subscription id, and the sku.  We no longer need to pass the plan information.

dvalinrh:
Has been fixed.

PR #155, opened 2025-01-21
Fix azure creation

dvalinrh:
# Description
It fixes Azure creation so it works again.

Azure has changed what it expects to be passed from terraform.  This updates the code to pass that info properly.

# Before/After Comparison
Before change:  Azure VM creations would fail.
After change: Azure VM creations succeed.

# Clerical Stuff
Issue: #154 
Relates to JIRA: RPOPC-306


kdvalin (approved):
LGTM

Issue #130, opened 2024-12-09
egrep deprecation needs cleanup


frival:
When running Zathras on newer RHEL or Fedora releases we get the following warning:

`egrep: warning: egrep is obsolescent; using grep -E`

We should probably make the conversion before the deprecation becomes a removal and things mysteriously break.

PR #156, opened 2025-01-22
Change egrep to grep -E

dvalinrh:
# Description
Change egrep to be egrep -E

# Before/After Comparison
We are getting this message while running burden
egrep: warning: egrep is obsolescent; using grep -E

This change removes that warning message.

# Clerical Stuff
Issue: #130

Relates to JIRA: RPOPC-273


frival (changes requested):
While the code looks good you forgot to fill out the PR template properly.  Fix that and we can probably get an approval.

kdvalin (approved):
LGTM

frival (approved):
LGTM, thanks

Issue #158, opened 2025-01-28
Fix CI errors


kdvalin:
# The problem
Currently the Issue tagging workflow is failing and triggering on issue comments.  This leads to unnecessary notifications and incorrect statuses for PRs.  See #156 as an example. 

# Expected Behaviour
The status check should not trigger on issue comments, and should work correctly when the PR has the appropriate structure.

# Actual Behaviour
The status check triggers on issue_comments, leading to failure.  For some reason #156 is failing even though it says it found a parent issue.

# Screenshots or Relevant Logs
[PR Failure](https://github.com/redhat-performance/zathras/actions/runs/13007703102/job/36278152021)

[Issue comment](https://github.com/redhat-performance/zathras/actions/runs/12993748129)

kdvalin:
Closing this out until we hit other problems

PR #159, opened 2025-01-28
Issue Labelling Bugfixes

kdvalin:
# Description
This PR fixes some errors with the Issue Labeling workflow.  For one, it removes the `issue_comment` debug trigger mistakenly committed.  Secondly it fixes a typo in a conditional causing it to never see that a PR has any related issues, causing failure.

# Before/After Comparison
## Before
[Issue Comment Errors](https://github.com/redhat-performance/zathras/actions/runs/12993748129)

[PR Failures](https://github.com/redhat-performance/zathras/pull/156/checks)

## After
Issue comment trigger is removed, so the workflow will not be triggered on issue comments (or PR comments).

The conditional has been updated, causing failure only when it cannot find a parent issue.

# Clerical Stuff
Issue: #158 
Relates to JIRA: RPOPC-315

### DIFF
--- a/.github/workflows/issue_tagging.yaml
+++ b/.github/workflows/issue_tagging.yaml
@@ -9,7 +9,6 @@ env:
   GH_TOKEN: ${{ github.token }}
 
 on:
-  issue_comment:
   pull_request_target:
     types:
       - review_requested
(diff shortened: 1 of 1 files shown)


frival (approved):
LGTM.  As discussed on slack, please add the group review label so this can be pushed immediately.

Issue #163, opened 2025-02-07
Upate configs to be latest test versions


dvalinrh:

dvalinrh:
Fixed.

PR #164, opened 2025-02-07
Update to latest test versions

dvalinrh:
# Description
Update various tests to the latest configuration.
Fix pig so it is using pig-wrapper

# Before/After Comparison
Older versions of tests being executed.

# Clerical Stuff
Issue:  #163 
Relates to JIRA: RPOPC-324

### DIFF
--- a/config/streams_template.yml
+++ b/config/streams_template.yml
@@ -1,6 +1,6 @@
 location: https://github.com/redhat-performance/streams-wrapper/archive/refs/tags
-exec_dir: "streams-wrapper-1.24/streams"
-repo_file: "v1.24.zip"
+exec_dir: "streams-wrapper-1.23/streams"
+repo_file: "v1.23.zip"
 os_supported: all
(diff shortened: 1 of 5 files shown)


frival (dismissed):
LGTM

config/streams_template.yml
```diff
@@ -1,6 +1,6 @@
 location: https://github.com/redhat-performance/streams-wrapper/archive/refs/tags
-exec_dir: "streams-wrapper-1.22/streams"
-repo_file: "v1.22.zip"
+exec_dir: "streams-wrapper-1.24/streams"
+repo_file: "v1.24.zip"
```
dvalinrh: streams 24 should to be 23.

dvalinrh: test version standard.
Fix uperf options

dvalinrh: Have resolved.

frival (approved):
The uperf change probably should really be a separate commit (since it's not test versions but config changes) but I'll let it go this time. :) LGTM

PR #168, opened 2025-02-11
Enable Reusable workflows

### LINKED ISSUE #167: Enable CI to be reusable
# The problem
Currently the GitHub workflows are only applicable to this repository.  If we want to scale it out to other repositories, we would need to copy them or make them reusable.

# Expected Behaviour
The workflow files can be edited from 1 repository and it'll propagate to all repositories.

# Actual Behaviour
If a workflow is edited, it will need to be copied off to all other wrapper repositories.

kdvalin:
# Description
This PR allows workflows defined within to be reusable.  This enables us to have a single point of configuration for workflows, instead of having to copy paste them throughout the repositories when CI changes.

# Before/After Comparison
## Before
CI would need to be copied and pasted to each repository, requiring a PR for each change in each repository.

## After
Workflow changes are restricted to this repository, leaving a single point of configuration

# Clerical Stuff
This closes #167 

Relates to JIRA: RPOPC-327

### DIFF
--- a/.github/workflows/check_review_label.yaml
+++ b/.github/workflows/check_review_label.yaml
@@ -1,6 +1,7 @@
 name: Verify group review
 
 on:
+  workflow_call:
   pull_request:
     types:
       - opened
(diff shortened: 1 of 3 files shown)


dvalinrh:
LGTM

dvalinrh (approved):
LGTM

PR #172, opened 2025-02-12
Update issue tagging to work without helper scripts

kdvalin:
# Description
This updates the issue tagging pipeline to work without relying on helper scripts.  This allows for a simple import into other repositories to use.

# Before/After Comparison
## Before
Importing the workflow into another repo would cause it to error out since the helper scripts are not present.

## After
Importing the workflow into another repository works as expected

# Clerical Stuff

Issue #167 

Relates to JIRA: RPOPC-327


tmp.json
```diff
@@ -0,0 +1,44 @@
+{
```
frival: Is this included by accident?  If not it could really use a better file name than "tmp.json" unless the workflow demands it.

frival (changes requested):
Looks like you left in a file by accident?

tmp.json
kdvalin: Whoops, had that as a testing file and forgot to remove it.

frival (dismissed):
LGTM.  Please file an issue to document the jira-issues-tagging container for other developers & maintainers.

dvalinrh (dismissed):
LGTM

dvalinrh (approved):
LGTM

PR #178, opened 2025-02-20
Remove perl package

### LINKED ISSUE #177: remove perl uuid package from streams
streams config contains perl uuid package, do not need it, remove

dvalinrh:
# Description
Removes loading the perl uuid package from the streams config

# Before/After Comparison
Before: streams would load perl uuid, which might fail, we do not use that package.
After: No longer loads that package.

# Clerical Stuff
This closes #177 

Relates to JIRA: RPOPC-349


kdvalin (approved):
LGTM

Issue #38, opened 2023-05-10
Zathras should not append to `<test>.cmd`


kdvalin:
Zathras currently appends to `<test>.cmd`.  When a Zathras scenario is run multiple times on the same host, old (potentially incorrect) options are still attempted, and multiple test runs can unintentionally occur.

dvalinrh:
Fixed

PR #182, opened 2025-02-25
CI typo hotfix

### LINKED ISSUE #181: Fix CI Container workflow
# The problem
The Jira CI container build workflow has a typo in it, preventing Quay login, and causing the workflow to error out.

# Expected Behaviour
The workflow should log into quay when a push to `main` happens.

# Actual Behaviour
A typo prevents any Quay login from occurring

kdvalin:
# Description
Fixes a typo in the Jira CI container so it can functionally work.

# Before/After Comparison
## Before
Workflows would skip the login step, preventing the ability to push new releases of the CI image.

## After
Workflows will now log in, and have the ability to push new releases of the CI image.

# Clerical Stuff
This closes #181 

Relates to JIRA: RPOPC-353

### DIFF
--- a/.github/workflows/build_ci_container.yaml
+++ b/.github/workflows/build_ci_container.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Login to Quay
-        if: github.event.name == 'push'
+        if: github.event_name == 'push'
         run: docker login quay.io -u ${{ secrets.QUAY_UNAME }} -p ${{ secrets.QUAY_PASSWD }}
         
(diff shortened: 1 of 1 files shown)


frival (approved):
LGTM.  Do we need to bypass the group review for this?

kdvalin:
This will prevent the container from building and pushing automatically.  There _might_ be something in there that can be used, but I don't recall.

If what's in there is non-functional, this blocks efforts for [pyperf-wrapper #38](https://github.com/redhat-performance/pyperf-wrapper/pull/38)

PR #184, opened 2025-03-04
Make parent issue script executable

### LINKED ISSUE #181: Fix CI Container workflow

kdvalin:
# Description
This makes a helper script executable, which is currently failing pipelines.

# Before/After Comparison
## Before
Script is not executable, pipelines fail as a result.

## After
Script is executable, pipelines no longer fail.

# Clerical Stuff
This closes #181

Relates to JIRA: RPOPC-353


dvalinrh (approved):
LGTM

PR #191, opened 2025-03-19
Install jq based on os vendor

### LINKED ISSUE #185: Ubuntu Failures on GCP
Ubuntu runs fails on GCP. Zathras tries to run 'yum' on ubuntu machine.

sousinha97:
# Description
Install jq based on the os vendor

This closes #185 
Relates to JIRA: RPOPC-340

### DIFF
--- a/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
+++ b/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
@@ -21,7 +21,7 @@
     - name: Install jq on RHEL
       when: config_info.os_vendor == 'rhel'
       shell: yum -y install jq
-
+    
     - name: Install jq on Ubuntu
       when: config_info.os_vendor == 'ubuntu'
       shell: apt-get update && apt-get install -y jq
--- a/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
+++ b/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
@@ -18,14 +18,6 @@
 
 - name: Retrieve the GCP meta data
   block:
-    - name: Install jq on RHEL
-      when: config_info.os_vendor == 'rhel'
-      shell: yum -y install jq
-    
-    - name: Install jq on Ubuntu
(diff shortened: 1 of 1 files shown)


dvalinrh (changes requested):
Do not need to add code to do this.  Update the test config files to include jq.  If jq is already installed it will simply return it is (just tried it on RHEL).

ansible_roles/roles/retrieve_meta_data/tasks/main.yml
```diff
@@ -16,17 +16,26 @@
...
-    shell: yum -y install jq
-  - name: Grab the metadata
-    shell: curl -H Authorization:Bearer $(gcloud auth print-identity-token) -H Metadata-Flavor:Google 'http://metadata.google.internal/computeMetadata/v1/instance/?recursive=true' | jq > /tmp/meta_data.yml
-    register: gcp_meta_data
-  - name: retrieve the gcp metadata
-    fetch:
-      src: "/tmp/meta_data.yml"
-      dest: "{{ working_dir }}/meta_data.yml"
-      flat: yes
+    - name: Install jq on RHEL
```
frival: Instead of doing this, shouldn't we be using the `package:` directive?  That's supposed to "transparently" work for whatever the underlying OS is.  The one concern with that is when we go to Image Mode we'll have to deal with any of these failing because you can't install new package in IM.

frival (changes requested):
Can we use package: instead, and contemplate image mode-related failures and how to mitigate.

ansible_roles/roles/retrieve_meta_data/tasks/main.yml
sousinha97: I tried it out. But if fails to work 

TASK [retrieve_meta_data : Install jq] *****************************************
fatal: [34.58.127.138]: FAILED! => {"changed": false, "msg": "No package matching 'jq' is available"}


@dvalinrh suggested to add package details in config files

dvalinrh:
Not doing at this time.

PR #197, opened 2025-03-27
Fix it so we only execute pbench_prep when we are installing pbench.

### LINKED ISSUE #196: ignore pbench prep when not installing pbench
Phase 1 of getting rid of pbench.

dvalinrh:
# Description
Stops Zathras from executing pbench prep if we are not installing pbench.

# Before/After Comparison
Before change, possible abort of execution of Zathras because failure in pbench prep, which should never have been called.

After change:  No longer execute pbench prep unless pbench is being installed.  No longer possible abort because of the failure.

# Clerical Stuff
This closes #196 

Relates to JIRA: RPOPC-388

### DIFF
--- a/bin/ten_of_us.yml
+++ b/bin/ten_of_us.yml
@@ -376,14 +376,17 @@
     - name: Gather the rpm package facts
       package_facts:
         manager: auto
-    - name: pbench prep
-      include_role:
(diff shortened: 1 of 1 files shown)


kdvalin (approved):
LGTM

PR #198, opened 2025-03-27
Fix leaving the zahtras gl lock around when updating the test versions.

### LINKED ISSUE #194: update_test_versions leaving things locked.
The option update test_versions is not releasing the exclusive lock.  Issue is the code for update is separate from the rest of burden, and we do remove the global lock directory when completed.  This locks out all further runs.

dvalinrh:
# Description
When we do an update_test_versions, we are not removing the global lock burden uses to protect pieces from multiple burdens running at the same time.

# Before/After Comparison
Before change, once you ran with --update_test_versions, any run of burden afterwards would hang trying to get the lock.

After the change, we now remove the lock after --update_test_versions is completed.  This allows ffuture runs of burden in the directory to actually run.

# Clerical Stuff
This closes #194 

Relates to JIRA: RPOPC-386

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -38,8 +38,8 @@
 #
 # For locking for blocks of shell code, we use
 #    Exclusive general lock
-#	To lock: gl_lock_exclusive <locking from, debug only>
-#	To unlock: gl_unlock_exclusive <unlocking from, debug only>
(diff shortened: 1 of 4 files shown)


kdvalin (approved):
LGTM

PR #199, opened 2025-03-27
We need to check to make sure the azure urn is valid.

### LINKED ISSUE #5: need to flag issue when no os provided to azure.
Flag issue when no version information is provided to azure.

dvalinrh:
# Description
Existing burden code does not check if the Azure urn is valid

# Before/After Comparison
Before the change, burden would proceed to trying to create the Azure instance.
After the change burden will catch the invalid urn and not attempt to create the Azure instance.

# Clerical Stuff
This closes #5 

Relates to JIRA: RPOPC-271


kdvalin (approved):
LGTM

PR #204, opened 2025-04-25
Do not attempt to Enable CodeReady Builder if we have said no packages.

### LINKED ISSUE #203: We are trying to load packages in, when designated not to.
There are a couple places where we are trying to load packages in, when we have been told not to.

dvalinrh:
# Description
Prevents doing anything with Enable CodeReady Builder if we are not to load packags.

# Before/After Comparison
Before:  We would attempt to do CodeReady Builder, when no_packages was designated.  This can cause a failure when checking the status file.  We do not want to perform this operation when no_packags is designated.

After:  We no longer attempt to do anything with CodeReady Builder, if no_packatges is designated.
# Clerical Stuff
This closes #203 

Relates to JIRA: RPOPC-414

### DIFF
--- a/bin/ten_of_us.yml
+++ b/bin/ten_of_us.yml
@@ -305,6 +305,7 @@
         - config_info.system_type != "local"
     when:
       - config_info.os_vendor == "rhel"
+      - config_info.do_not_install_packages == 0
 
 - hosts: local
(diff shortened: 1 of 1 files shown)


grdumas (approved):
LGTM

PR #206, opened 2025-05-05
update iozone options for auto mode

### LINKED ISSUE #205: Iozone default options should be updted since auto mode works and throughput mode has issues
Problem: Now that the iozone wrapper properly supports "automatic mode" at or near parity with the old automation the default options should be updated to use it. The current defaults are left over from the initial drop of the iozone wrapper and use the (currently) less useful throughput mode in a minimalist way.

Current: The defaults for the iozone wrapper are for a minimalist throughput mode 
Expected: The defaults set us up for successful test sweeps in automatic mode

malucius-rh:
# Description
Now that auto mode is working and usable and throughput mode is less useful for regression update the default options

# Before/After Comparison
Before: The iozone wrapper sets up runs for a couple of subtests run in throughput mode
After: The iozone wrapper set up runs to mimic the legacy automoation: automatic mode, all I/O types, all subtests, three filesystems

# Clerical Stuff
This closes #205 
Relates to JIRA: RPOPC-433

### DIFF
--- a/config/test_defs.yml
+++ b/config/test_defs.yml
@@ -176,8 +176,7 @@ test_defs:
     test_description: iozone
     pbench_required: "no"
     pbench_local_results: "no"
-    test_specific: "--devices_to_use {{ dyn_data.storage }} --filesys xfs --file_count_list 1,2,4 --test_type 0,1 --incache --mount_location /iozone/iozone0"
-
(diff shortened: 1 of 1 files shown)


kdvalin (approved):
LGTM

PR #214, opened 2025-05-20
Allow for Azure OS designation to happen via URN as well as name.

### LINKED ISSUE #202: Need to allow use of subscription id for azure images.
We need to add code to handle subscription based images for azure.  With out this change we are unable to access privately built images.

dvalinrh:
# Description
Allows the designation of the OS for Azure to use to be either name or a URN.  This is required for use of private images. If a name is provided for the OS, it will be converted to a URN.

# Before/After Comparison
Before, we were not able to use a private images
After, we can use a private images.

# Clerical Stuff
This closes #202

Relates to JIRA: RPOPC-411


ansible_roles/roles/azure_create/tasks/main.yml
```diff
@@ -11,12 +11,24 @@
...
+    when: config_info.cloud_os_version != "none"
+  - name: Using subscription for image.
+    block:
+    - name: cat non spot machine terraform
+      shell: "cat {{ config_info.local_run_dir }}/ansible_roles/roles/azure_create/files/tf/vm_spot_set_sub.tf | sed \"s/PRIORITYSPOT//g\" | sed \"s/EVICTIONPOLICY//g\" >> {{ working_dir }}/tf/main.tf" 
+      when: config_info.spot_start_price == 0
+    - name: cat spot machine terraform
+      shell: "cat {{ config_info.local_run_dir }}/ansible_roles/roles/azure_create/files/tf/vm_spot_set_sub.tf | sed \"s/PRIORITYSPOT/priority              = \\\"Spot\\\"/g\" | sed \"s/EVICTIONPOLICY/eviction_policy       = \\\"Deallocate\\\"/g\" >> {{ working_dir }}/tf/main.tf" 
+      when: config_info.spot_start_price == 1
+    when: config_info.cloud_os_version == "none"
```
kdvalin, line 31: At some point we might want to consider using jinja2 templates for this..

This isn't really readable.

bin/ten_of_us.yml
```diff
@@ -234,6 +234,7 @@
           status_file: "/tmp/cr_status"
       when:
         - config_info.system_type == "azure"
+        - config_info.do_not_install_packages == 0
```
kdvalin, line 237: Will start nitting more and more on fixing multiple things in one commit.

Commits are cheap, we should use them as essentially code checkpoints.

kdvalin (approved):
Some general overall nits.  LGTM otherwise

PR #215, opened 2025-05-20
Adding support for ctrl c

### LINKED ISSUE #210: Catch signals in burden, and bail out accordingly
Right now, if for some reason we break out of burden, there is a lot of files being left around, including the lock files.  This pr fixes this issue, by trapping the signal and bailing out.  Note, as in the case of sig 9, there is no way to catch it.

dvalinrh:
# Description
This adds code to trap a ctrl c from the user and then cleanup accordingly.

# Before/After Comparison
Before:  A ctrl c would break from the first level of burden, but leave the ansible playbook running, as well as the lock files and any cloud system created.
After:  The ctrl c is now caught and handled accordingly.  A full clean up of the environment is done, as well as termination of the cloud system if we are suppose to terminate the system at completion.

# Clerical Stuff
This closes #210 

Relates to JIRA: RPOPC-443


dvalinrh:
Have tested with cloud (azure and aws, code is the same for all clouds), as well as bare metal.  Lock files are cleaned up properly, and the clouds terminated if requested to do so.

bin/burden
```diff
@@ -312,13 +317,31 @@ cleanup_and_exit()
...
+	fi
 	if [[ $gl_scenario_to_restore != "" ]]; then
 		scenario_restore="--restore_file $gl_scenario_to_restore,$gl_scenario_to_run"
 	fi
 	if [[ $3 != "" ]]; then
 		sysname="--sysname $3"
 	fi
 	cd $gl_top_dir
+
+	if [[ $gl_run_pindex -eq $gl_max_systems ]]; then
```
dvalinrh: Wrong.

bin/utils/cleanup_and_exit_out_back
```diff
@@ -0,0 +1,199 @@
+#!/bin/bash
```
dvalinrh: Delete file.

config/test_defs.yml
```diff
@@ -176,7 +176,7 @@ test_defs:
     test_description: iozone
     pbench_required: "no"
     pbench_local_results: "no"
-    test_specific: "--devices_to_use {{ dyn_data.storage }} "--filesys xfs,ext4,ext3  --all_test --mount_location /iozone/iozone0 --eatmem --auto"
+    test_specific: "--devices_to_use {{ dyn_data.storage }} --filesys xfs,ext4,ext3  --all_test --mount_location /iozone/iozone0 --eatmem --auto"
```
dvalinrh, line 179: Already fixed.

bin/utils/terraform_ops
```diff
@@ -35,24 +35,29 @@ tf_terminate()
...
+			fi
 			#
-			# We do not want to exit out on an error.  The instance could have been
-			# deleted via the web, and there are other instances to
-			# delete.
+			# Get arid of the tf directory
 			#
-			echo "Warning: Unable to create the plan for $tf_del"
+			cd ..
+#			rm -rf tf
```
kdvalin:
```suggestion
			rm -rf tf
```

kdvalin (changes requested):
Verify functionality of keeping clouds as well.

kdvalin:
Missing Jira Ticket.

Did you verify that it won't clean up cloud systems if told to?

dvalinrh:
Verified, cloud system remains if breaking out and told not to terminate results upon completion.  Verified cloud system is terminated if told to do so and we break out.

kdvalin (approved):
LGTM

PR #224, opened 2025-06-06
Remove data_gather, script has been moved to test_tools-wrapper

### LINKED ISSUE #223: Remove data_gather.
data_gather has been move to test_tools-wrapper, and is known as gather_data.

dvalinrh:
# Description
Completes the move of data_gather to test_tools-wrapper and is known as gather_data.

# Before/After Comparison
Before this change the test wrappers may fail to gather the system data as it we are expecting zathras to be installed, which might not be the case.

After the change, the wrappers will be able to gather system information regardless if using Zathras or not.  Note, the system data is now only stored with the wrapper.

# Clerical Stuff
This closes #223 

Relates to JIRA: RPOPC-474


kdvalin (approved):
LGTM

PR #231, opened 2025-06-18
Add error handling for unsupported systems

### LINKED ISSUE #232: Zathras installer script should check if user is running on a supported system
### Issue
Currently there are no plans to bring support for the Zathras installer script beyond Fedora and RHEL. Rather than rewrite the script to support multiple package managers, it would be simpler to have the installer immediately fail and inform the user to either install on a DNF-based system or manually perform the setup steps.

Sidebar: how many changes to Zathras would be required to support other major distros (Ubuntu, maybe Debian)? Maybe someday worth exploring. 

### Current behavior:
Install script will proceed through setup steps until hitting the first use `dnf` and failing.

### Desired behavior:
Install script should immediately fail if the installation destination is not DNF-based.

grdumas:
# Description
Currently Zathras installation support is only planned for Fedora and RHEL. This PR adds error handling for unsupported systems in two ways.

First, it requires the installation destination to have the DNF package manager. DNF is required by subsequent installation steps, and if not present, will fail quickly rather than attempting some setup steps.

Second, it cleans up the logic around Terraform installation so that installation on unsupported systems results in a clear error message.

The code changes were created with assistance from Cursor, using Claude AI. Testing was performed by a human. 

This is some of the first AI-assisted code being introduced into the project - let's examine and discuss its inclusion carefully. Should there be a different process for AI generated or assisted code? Additional testing steps? Labels - on commits, PRs?

# Before/After Comparison
Tested on a system with DNF package manager (CentOS) and a non-DNF system (Ubuntu).

### Before
```
tester@ubuntu:~/zathras $./bin/install.sh
Not running as root, proceed.
Installing ansible-core...
[sudo] password for tester:
sudo: dnf: command not found
```

### After
```
tester@ubuntu:~/zathras $./bin/install.sh
Error: This script requires the DNF package manager.
Supported distributions: RHEL 8+, Fedora, CentOS 8+, Rocky Linux, AlmaLinux
For other distributions, please install the required packages manually:
  - ansible-core, git, jq, python, python3-pip, terraform, unzip, wget
```

### Before:
Installation begins without error, but fails to install Terraform without clear error message.
```
...
python-pip3 is installed.
./bin/install.sh: line 62: release: unbound variable
```

### After:
Installation begins without error, fails on Terraform install with clear error message.
```
...
python3-pip is installed.
Error: Terraform installation is only supported on RHEL and Fedora distributions.
Please install Terraform manually from https://developer.hashicorp.com/terraform/install
```

### Before
Install works as expected.

### After
Install works as expected.

# Clerical Stuff
Closes #232 

Relates to JIRA: RPOPC-498

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -57,6 +57,11 @@ for package in "${packages[@]}"; do
             release='RHEL'
         elif [ $os_release_clean = 'fedora' ]; then
             release='fedora'
+        else
+            echo "Error: Terraform installation is only supported on RHEL and Fedora distributions."
+            echo "Detected OS: $os_release_clean"
+            echo "Please install Terraform manually from https://developer.hashicorp.com/terraform/install"
+            exit 1
         fi
             # repo URL for terraform
             repo_url="https://rpm.releases.hashicorp.com/${release}/hashicorp.repo"
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -28,7 +28,7 @@ set -eu
 # - Test brackets are only needed for comparisons like [[ "$a" == "$b" ]] or [[ -f "$file" ]]
 if ! command -v dnf &> /dev/null; then
     echo "Error: This script requires the DNF package manager."
-    echo "Supported distributions: RHEL 8+, Fedora, CentOS 8+, Rocky Linux, AlmaLinux"
+    echo "Installer supported distributions: RHEL 8+, Fedora"
     echo "For other distributions, please install the required packages manually:"
     echo "  - ansible-core, git, jq, python, python3-pip, terraform, unzip, wget"
     exit 1
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -63,14 +63,15 @@ for package in "${packages[@]}"; do
             echo "Please install Terraform manually from https://developer.hashicorp.com/terraform/install"
             exit 1
         fi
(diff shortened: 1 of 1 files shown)


bin/install.sh
```diff
@@ -19,6 +19,15 @@
 
 set -eu
 
+# Check if dnf package manager is available
+if ! command -v dnf &> /dev/null; then
```
dvalinrh, line 29: Add [[ ]]

bin/install.sh
```diff
@@ -19,6 +19,15 @@
...
+    echo "Error: This script requires the DNF package manager."
+    echo "Supported distributions: RHEL 8+, Fedora, CentOS 8+, Rocky Linux, AlmaLinux"
```
kdvalin: CentOS/Rocky/Alma are not supported by this script

kdvalin (changes requested):
Fix supported OS's

bin/install.sh
grdumas: Added comments explaining why test brackets aren't needed.

bin/install.sh
grdumas: Removed everything after RHEL and Fedora, marking fixed.

kdvalin (approved):
LGTM

Issue #238, opened 2025-07-14
Add PullHero for first pass AI PR reviews


kdvalin:
We should add PullHero for first pass code reviews on PRs.

This should help contributors by catching obvious problems quickly and without requiring a maintainer to look over the code.

Requirements:
- Automatically trigger code review when a pr is opened
- Trigger a code review with a `/review` comment from maintainers
- Trigger a code review when a review is requested
- The code review provided by AI should not count as an approval or a request for changes, requiring a maintainer to give final say

PR #240, opened 2025-07-28
Fix PullHero PR Comment Trigger

kdvalin:
# Description
This PR fixes the PullHero `/review` comment trigger.

# Before/After Comparison
## Before
PullHero would error out due to an inability to fetch the correct refs from the enviornment.

## After
PullHero should work successfully

# Clerical Stuff
Continues work on #238 
Relates to JIRA: RPOPC-533


grdumas (approved):
LGTM

PR #243, opened 2025-07-30
Check to make sure we are not running too new of a version of yq.

### LINKED ISSUE #242: Version check yq
Zathras depends on the yq version being 2.10.0.  If it is not, things will fail in unpredictable ways.  To resolve this, a check will be add for the version number of yq,and if it is not proper, we will abort.

dvalinrh:
# Description
Makes sure the version of yq we are running is a version Zathras can use.
If the version is too new, Zathras will fail in unpredictable ways.

# Before/After Comparison
Before Change: If the version of yq is too new, Zathras will fail unpredictably. 
After Change: If the version of yq is too new, Zathras will bail with an error saying so.

# Clerical Stuff
This closes #242 

Relates to JIRA: RPOPC-544

# Testing
Verified the versioning  works by passing the following
1) Version less than the max allowed.  Works
2) Version that is the max allowed.  Works
3) Version greater than the max allowed.  Failed as expected.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -2213,6 +2213,11 @@ report_util_version()
 	echo ${1}: ${2} >> utils_version
 }
 
+version_greater_equal()
+{
+	printf '%s\n%s\n' "$2" "$1" | sort --reverse --version-sort | head -n 1
+}
+
 #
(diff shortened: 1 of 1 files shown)


grdumas (approved):
LGTM.

PR #245, opened 2025-07-31
Fix issues with creating azure systems.

### LINKED ISSUE #244: azure image creation issues.
Two issues:
1) Azure has increased the min. disk size to 64GB
2) Code ready repo name has changed.

To fix # 2 we will use the logic in place for AWS, works perfectly fine.

dvalinrh:
# Description
Fixes the creation of azure images
# Before/After Comparison
Before: We would fail because the root disk size is too small.  Once fixed, we failed because the code readyrepo name changed.
After:  Creation of azure systems work fine.

# Clerical Stuff
This closes #244 

Relates to JIRA: RPOPC-545
# Testing
Verified that the azure systems are created properly.


kdvalin (approved):
LGTM

PR #247, opened 2025-07-31
Removing pbench support from Zathras.

### LINKED ISSUE #246: Remove pbench support
pbench product has been EOL'd and no longer runs on RHEL 10.  Remove code in preparation for pcp.

dvalinrh:
# Description
Removes the pbench infrastructure from Zathras.  PCP will be replacing pbench, as pbench no longer is supported and does not run on RHEL 10.

# Before/After Comparison
Before: Attempt to use pbench behavior at best fails, at worst will be unpredictable.
After: pbench support has been completely removed.  Any scenario file attempting to reference pbench, will fail at startup of burden.

# Clerical Stuff
This closes #246 

Relates to JIRA: RPOPC-546

# Testing
Verified local systems runs still work.
Verified AWS system creation still works.
Verified Azure system creation still works.
GCP did not have any pbench code embedded in the creation role.


config/failed_test
```diff
@@ -0,0 +1,3 @@
+Thu Jul 31 08:55:30 AM EDT 2025 n42-h30-000-r650.rdu3.labs.perfscale.redhat.com, Only github is supported.
```
kdvalin: Remove auto generated file

dvalinrh: Delete file.

kdvalin (changes requested):
Otherwise, LGTM

kdvalin (approved):
LGTM

PR #251, opened 2025-08-01
Remove unneeded line.

### LINKED ISSUE #250: Remove unneeded line from pig template.
The line in the pig config template
pre_setup: /home/dvalin/install_bc

Should not be there, removing.

dvalinrh:
# Description
Remove an old entry from the pig config template

# Before/After Comparison
Before: Line is in template that should not be there.
After: Line is no longer in template
No change in behavior, just removing a line no longer required, and may cause issues in the future.

# Clerical Stuff
This closes #250 

Relates to JIRA: RPOPC-550


kdvalin (approved):
LGTM

PR #253, opened 2025-08-04
Provide `--no_pkg_install` to wrappers when needed

### LINKED ISSUE #252: `--no_pkg_install` is not passed during `--no_packages` runs
# The problem
When passing `--no_packages` to burden, the wrappers should also receive a flag to skip installing packages.  This is presently not being done.  This can lead to unexpected package installs, or straight up errors in the case of bootc hosts.

The ansible parts do skip over package install, but if the wrapper does any amount of package installations it will not be skipped.

# Workarounds
None currently.

kdvalin:
# Description
This PR adds `--no_pkg_install`  to wrappers when package installation would be skipped otherwise.

# Before/After Comparison
## Before
`./burden --no_packages` would not give `--no_pkg_install` to the running wrappers

## After
`./burden --no_packages` will give `--no_pkg_install` to the running wrapper.

# Clerical Stuff
This should fix #252

Relates to JIRA: RPOPC-551

### DIFF
--- a/ansible_roles/roles/test_generic/tasks/main.yml
+++ b/ansible_roles/roles/test_generic/tasks/main.yml
@@ -79,6 +79,7 @@
         --host_config "{{ config_info.host_config_info }}"
         --sysname "{{ config_info.host_or_cloud_inst }}"
         --sys_type {{ config_info.system_type }}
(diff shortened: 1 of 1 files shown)


dvalinrh (approved):

grdumas (approved):
lgtm

PR #260, opened 2025-08-04
Error checking added to parse_disk_line

### LINKED ISSUE #259: Add error handling to parse_disk_line
parse_disk_line currently does not report errors on fields that are designated but 
are not valid.  Need to flag them and return an error code.

dvalinrh:
# Description
Hardens parse_disk_line, so it will detect fields that wrong.

# Before/After Comparison
Before:  If throughput field name was misspelled, we would simply ignore it, and assign the default value for throughput.
After:  If we can not determine the designated field, we flag it and return a 1 for error code.  If all fields are present, we will return a 0 for a rtc.

# Clerical Stuff
This closes #259 

Relates to JIRA: RPOPC-554

# Test
1) Verified that a bad field in the string passed to parse_disk_line is caught and return code of 1 is returned.
2) Verified that a valid string passed into parse_disk_line passes properly, and rtc of 0 is returned.
3) Verified the above using burden.

### DIFF
--- a/bin/parse_disk_line
+++ b/bin/parse_disk_line
@@ -33,7 +33,7 @@ set_disk_field()
 	elif [[ $dfield == *"throughput"* ]]; then
 		disk_tp=$value
 	else
-		echo $dfield: Unknown field, bail out.
+		echo echo "$dfield: Unknown field, bail out."
 		exit 1
 	fi
 }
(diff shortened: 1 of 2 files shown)


bin/parse_disk_line
```diff
@@ -45,28 +44,35 @@ set_disk_field()
 #
 instance_type=`echo $1 | cut -d: -f1`
 fields=`echo "$1" | sed "s/&/ /g"`
-
+index=0
 for i in $fields
 do
+	let "indecx=$index+1"
```
kdvalin: typo

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #266, opened 2025-08-06
Change systen type from aws to amazon.

### LINKED ISSUE #265: Change system type and os vendor from aws to be amazon
Naming change to clean things up a bit, and make it easier for filtering scripts.

dvalinrh:
# Description
We want the results are to contain amazon for the cloud name, not aws.

# Before/After Comparison
Before: results would be  test/rhel/aws
After: results are now test/rhel/amazon

# Clerical Stuff
This closes #265 

Relates to JIRA: RPOPC-557

# Tested
1) Verified aws systems still work.
2) Verified results directory is as expected.
3) Verified bare metal systems still worked.
4) Verfied azure systems still worked.

### DIFF
--- a/README
+++ b/README
@@ -45,7 +45,7 @@ General options
   --dir_index: designates an index value to add to the run directory, default is 0
   --force_upload: upload all packages and tools to the cloud instance.
   --git_timeout: Number of seconds to timeout on git requests.  Default is 60
-  --system_type <vendor>: aws, azure or local
+  --system_type <vendor>: amazon, azure or local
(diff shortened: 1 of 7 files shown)


kdvalin:
This seems like you missed the original ask, it should be that the results directories for Amazon Linux runs should be `amazon` , like RHEL is `rhel` and Ubuntu is `ubuntu`.

This shouldn't have anything to do with selecting whether or not AWS was used for the run.

dvalinrh:
Not required.

PR #277, opened 2025-08-27
Fix versioning so we do not include pre-releases as the latest version

### LINKED ISSUE #276: Do not include pre-release versions of test wrappers when setting the latest version.
Exclude pre-releases as being candidate for the latest test wrapper version.

dvalinrh:
# Description
Prevents pulling in pre-releases when we have asked to update the test versions

# Before/After Comparison
Before: We could pull in a pre-release version as an official version.
After: We will pull in the version that is marked as the latest.

# Clerical Stuff
This closes #276 

Relates to JIRA: RPOPC-584

# Testing
Verified the pre-release versions do not turn up.
Verified still can use a pre-release version.
Verfiied when updating the versions, the pre-release versions are not considered.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -51,6 +51,8 @@ else
 fi
 
 # check for and install system packages
+# We expect EPEL to be installed, which contains gh.
+#
 packages=(ansible-core git jq python python3-pip terraform unzip wget)
 
 for package in "${packages[@]}"; do 
--- a/bin/utils/test_versions_ops
+++ b/bin/utils/test_versions_ops
@@ -67,7 +67,8 @@ test_version_ops()
 		location=`grep location: ${test_repo} | cut -d: -f2- | sed "s/ //g"`
 		if [[ $location == "https://github.com"* ]]; then
 			repo=`echo $location | cut -d'/' -f 1-5`
-			git -c 'versionsort.suffix=-' ls-remote --exit-code --refs --sort='version:refname' --tags $repo | rev | cut -d'/' -f1 | rev >  git_vers
+			gh release list  --repo $repo | rev | awk '{print $1" "$2" "$3}' | rev > vers_file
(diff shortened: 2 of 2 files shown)


kdvalin (changes requested):
Update install script with `gh`

bin/install.sh
```diff
@@ -51,7 +51,7 @@ else
 fi
 
 # check for and install system packages
-packages=(ansible-core git jq python python3-pip terraform unzip wget)
+packages=(ansible-core git jq python python3-pip terraform unzip wget gh)
```
kdvalin: `gh` is not available in default RHEL repos, see https://github.com/cli/cli/blob/trunk/docs/install_linux.md#rpm

kdvalin: It is available in EPEL though

dvalinrh: Available several means, including epel

Last metadata expiration check: 2:06:20 ago on Tue 02 Sep 2025 10:01:59 AM EDT.
gh-2.65.0-1.x86_64 : GitHub’s official command line tool.
Repo        : @System
Matched from:
Filename    : /usr/bin/gh

gh-2.76.1-1.el9.x86_64 : GitHub's official command line tool
Repo        : epel
Matched from:
Filename    : /usr/bin/gh
Filename    : /usr/share/bash-completion/completions/gh
Filename    : /usr/share/doc/gh
Filename    : /usr/share/licenses/gh

gh-2.78.0-1.aarch64 : GitHub’s official command line tool.
Repo        : gh-cli
Matched from:
Filename    : /usr/bin/gh
Filename    : /usr/share/bash-completion/completions/gh

kdvalin: The `@System` means it was installed.

The `gh-cli` repo is an added repo, but not shipped by default on RHEL.

If we need EPEL we should document that somewhere (or install the repo automagically)

kdvalin (approved):
LGTM

PR #281, opened 2025-08-28
Remove the "2" file from showing up.

### LINKED ISSUE #279: Getting a bogus file "2" due to bad redirect.
Zathras is creating a "2" file due to bad redirects

grep ${look_for}: $tmpfile >2 /dev/null

This case we can do a grep -q

there are others to be fixed.

dvalinrh:
# Description
Remove a bogus file in the run directory

# Before/After Comparison
Before: A file named 2 would appear in the run directory
After: File no longer appears

# Clerical Stuff
This closes #279 

Relates to JIRA: RPOPC-588

# Test
Ran burden, still works, and we do not get a 2 file in the run directory.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -3096,7 +3096,7 @@ process_test_template_data()
 			while IFS= read -r fline
 			do
 				look_for=`echo $fline | cut -d: -f 1`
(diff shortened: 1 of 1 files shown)


kdvalin (approved):
LGTM

PR #285, opened 2025-09-02
First phase of adding Suse Enterprise Linux support.

### LINKED ISSUE #278: Add Suse Enterprise Linux support
Add Suse Enterprise Linux (SLES) to the supported target operating systems.  We'll want more than one (but not necessarily all) representative workloads supported for the initial push.  Don't worry about supporting SLES as a Zathras controller system for now either.

frival:
This has been tested on Azure so far, with testing on AWS, GCP, and bare metal pending.  This commit only includes workload configs that have been tested; others are either still under test or need work in the wrapper code proper.

Assisted-by: Claude Code

# Description
Adds support for SLES as a target OS as well as some config files for benchmarks that are tested to work.  This should be sufficient as a proof that the SLES-related code is reasonably correct for a first phase.

# Before/After Comparison
Before: SLES is not an option.
After: SLES is one of the options (hereafter named "suse")

# Clerical Stuff
Mention the issue this PR works toward resolving.  If the 
PR completely solves the issue, use "This closes #x"

This closes #278 
Relates to JIRA: RPOPC-586

### DIFF
--- a/ansible_roles/roles/install_packages/tasks/main.yml
+++ b/ansible_roles/roles/install_packages/tasks/main.yml
@@ -58,3 +58,37 @@
       rtc_val: "{{ install_results.rc }}"
       packages: "{{ pkg_install }}"
   when: config_info.os_vendor == "ubuntu"
+
+#
+# Install packages on SUSE images.
+#
+- name: install SUSE packages
+  block:
+  - name: Remove any existing status file.
+    shell: "rm -f {{ results_file }}"
+  - name: install zypper packages
+    zypper:
+      name: "{{ pkg_install }}"
+      state: latest
+    register: install_results
+    until: install_results is succeeded
+    retries: 5
+    delay: 10
+  - name: Report Success
+    lineinfile:
+       path: "{{ results_file }}"
+       line: "status: success"
+       create: yes
+  - name: Log package list
+    lineinfile:
+      path: "/tmp/frival_dbg"
+      line: "{{ pkg_install }}"
+      create: yes
+  rescue:
+    - name: Report Error
+      lineinfile:
+         path: "{{ results_file }}"
+         line: "status: failed"
+         create: yes
+  any_errors_fatal: true
+  when: config_info.os_vendor == "suse"
--- a/ansible_roles/roles/install_packages/tasks/main.yml
+++ b/ansible_roles/roles/install_packages/tasks/main.yml
@@ -79,11 +79,6 @@
        path: "{{ results_file }}"
        line: "status: success"
        create: yes
-  - name: Log package list
-    lineinfile:
-      path: "/tmp/frival_dbg"
-      line: "{{ pkg_install }}"
-      create: yes
   rescue:
     - name: Report Error
       lineinfile:
--- a/ansible_roles/roles/install_packages/tasks/main.yml
+++ b/ansible_roles/roles/install_packages/tasks/main.yml
@@ -62,6 +62,8 @@
 #
 # Install packages on SUSE images.
 #
+# The zypper module comes from community.general.  If at any point we find a case it's not already installed we'll have to update the installer script.
+# 
 - name: install SUSE packages
   block:
   - name: Remove any existing status file.
--- a/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
+++ b/ansible_roles/roles/retrieve_meta_data/tasks/main.yml
@@ -5,6 +5,14 @@
 #
 - name: Retrieve the azure meta data and set the tags
   block:
+  - name: Install jq
+    ansible.builtin.package:
+      name: jq
+      state: latest
+    register: jq_installed
+    until: jq_installed is succeeded
+    retries: 5
+    delay: 10
   - name: Grab the metadata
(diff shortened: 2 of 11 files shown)


ansible_roles/roles/install_packages/tasks/main.yml
```diff
@@ -58,3 +58,37 @@
...
+  - name: Report Success
+    lineinfile:
+       path: "{{ results_file }}"
+       line: "status: success"
+       create: yes
+  - name: Log package list
+    lineinfile:
+      path: "/tmp/frival_dbg"
+      line: "{{ pkg_install }}"
+      create: yes
```
dvalinrh: Delete debug

ansible_roles/roles/retrieve_meta_data/tasks/main.yml
```diff
@@ -5,6 +5,14 @@
...
   block:
+  - name: Install jq
+    ansible.builtin.package:
+      name: jq
+      state: latest
+    register: jq_installed
+    until: jq_installed is succeeded
+    retries: 5
+    delay: 10
   - name: Grab the metadata
```
dvalinrh, line 16: Handle failure.

bin/ten_of_us.yml
```diff
@@ -303,6 +303,49 @@
...
+  become: true
+  vars_files: ansible_vars.yml
+  gather_facts: no
+  tasks:
+  - name: enable SLES legacy module
+    block:
+    - name: Enable sle-module-legacy/15.7/x86_64
+      shell: "SUSEConnect -p sle-module-legacy/15.7/x86_64"
+      register: suse_connect_result
+      ignore_errors: yes
```
dvalinrh, line 321: what about ARM?

bin/ten_of_us.yml
```diff
@@ -303,6 +303,49 @@
...
+      include_role:
+        name: record_status
+      vars:
+        results: "{{ suse_connect_result }}"
+        status_file: "/tmp/sles_legacy_status"
+    - name: retrieve sles legacy module status
+      include_role:
+        name: retrieve_status
+      vars:
+        status_file: "sles_legacy_status"
```
dvalinrh, line 332: Add rescue code.

ansible_roles/roles/install_packages/tasks/main.yml
```diff
@@ -58,3 +58,37 @@
...
+
+#
+# Install packages on SUSE images.
+#
+- name: install SUSE packages
+  block:
+  - name: Remove any existing status file.
+    shell: "rm -f {{ results_file }}"
+  - name: install zypper packages
+    zypper:
```
kdvalin, line 72: Where is the zypper module available?  It seems like it's from `community.general`, which would need to be added to the install script

bin/ten_of_us.yml
```diff
@@ -303,6 +303,49 @@
...
+  user: config_info.test_user
+  become_user: root
+  become: true
+  vars_files: ansible_vars.yml
+  gather_facts: no
+  tasks:
+  - name: enable SLES legacy module
+    block:
+    - name: Enable sle-module-legacy/15.7/x86_64
+      shell: "SUSEConnect -p sle-module-legacy/15.7/x86_64"
```
kdvalin: Is 15.7 the \<Release\>.\<Service pack\>?  If so this shouldn't be hardcoded.

kdvalin (changes requested):
Couple nits

ansible_roles/roles/install_packages/tasks/main.yml
frival: Done

bin/ten_of_us.yml
frival: Replaced x86_64 with "{{ ansible_architecture }}" and verified this gives us the correct module still.

ansible_roles/roles/install_packages/tasks/main.yml
frival: This is running on the remote host not the Zathras controller so it wouldn't belong in the install script.  That said, I didn't have to install anything for this to work on the hosts I've tried so far.  I'll try to find a config where this fails and if I can then I should be able to add an early step to install the collection with ansible-galaxy.  I'm just a little leery of adding code I'm unable to test.

bin/ten_of_us.yml
frival: Replaced 15.7 with "{{ ansible_distribution_version }}" which looks to be doing what we're looking for.

bin/ten_of_us.yml
frival: Added a rescue stanza similar to others.

ansible_roles/roles/install_packages/tasks/main.yml
frival: Added a comment for the future per our offline discussion.

ansible_roles/roles/retrieve_meta_data/tasks/main.yml
frival: Rescue clause added per offline discussion.

dvalinrh (approved):

kdvalin (approved):
LGTM

PR #289, opened 2025-10-02
Fix it so the private network stays with ubuntu.

### LINKED ISSUE #288: ubuntu is dropping the private network.
Due to the reboot and netplan, we lose the private network with ubuntu.  This
is causing ssh_key_exchange to fail.

dvalinrh:
# Description
Fixes it so the private network ip is present when we run uperf.

# Before/After Comparison
Before:  Due to the reboot, and restarting ssh we lose the ipv4 private ip address.
After:  Address is present.

# Clerical Stuff
This closes #288 

Relates to JIRA: RPOPC-636


dvalinrh:
Closed, ubuntu is working fine now.

PR #291, opened 2025-10-03
Fix it so azure does not run uperf on both systems.

### LINKED ISSUE #290: azure is running uperf on both client and server.
We are adding both client and server to test_hosts.

dvalinrh:
# Description
Fixes it so we are not running uperf on both client/server in the azure environment.

# Before/After Comparison
Before: When using azure, uperf is being executed on both the client and server.
# Clerical Stuff
This closes #290 

Relates to JIRA: RPOPC-637


frival (approved):
LGTM

PR #297, opened 2025-10-08
Remove check for python3.

### LINKED ISSUE #296: Remove check for python3 on aws system creation
This code is no longer required, and dates back to when we had issues of python3 not being installed.

dvalinrh:
# Description
Removes the check for python3 when creating aws systems

# Before/After Comparison
Before:  Check to see if python is installed, if not bail.  This check dates back to when we started original cloud work, and python may not be installed.  Python is installed now on the initial bringup and we are failing at times because we are checking for a specific python.
After: Check is no longer done.
# Clerical Stuff
This closes #296 

Relates to JIRA: RPOPC-643


dvalinrh:
verified we do not say we support RHEL 8.

kdvalin (approved):
LGTM

PR #303, opened 2025-10-23
Replace zip from git with tar.gz

### LINKED ISSUE #302: change using the git file *.zip  to .tar.gz
With the moving to packages, zip may not always be installed on the various OS flavors.  Instead use .tar.gz

dvalinrh:
# Description
Having issues with some linux derivatives not having zip present, and will cause a chicken egg issue with the packaging changes coming.  This switches to use tar.gz,  all derivatives of linux tried on, are happy.

# Before/After Comparison
Before: Zathras would fail on some flavors of linux if unzip package was not brought in
After: Zathras is happy with all flavors now, unzip package is not required.

# Clerical Stuff
This closes #302 

Relates to JIRA: RPOPC-659

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -114,6 +114,7 @@ gl_retry_failed_tests=1
 gl_base_cost=0
 gl_host_uses_nvme=0
 gl_cloud_net_type="default"
+gl_version_check_only=0
 #
 # We will default to eastus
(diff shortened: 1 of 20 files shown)


kdvalin (approved):
LGTM

PR #321, opened 2025-11-11
Fix it so /uploads exists.

### LINKED ISSUE #320: upload of tar balls is broken.
We no longer have /uploads on the created systems.  /uploads is where the various tests that requires uploaded kits expect to find the kits.

dvalinrh:
# Description
Restores the existence of /uploads

# Before/After Comparison
Before:  Uploads of kits occured, but the common location /uploads is missing.
After: /uploads is present again, allowing the execution of wrappers requiring uploaded kits.

# Documentation Check
No
# Clerical Stuff
This closes #320

Relates to JIRA: RPOPC-698

### DIFF
--- a/ansible_roles/roles/upload_extra/tasks/main.yml
+++ b/ansible_roles/roles/upload_extra/tasks/main.yml
@@ -27,3 +27,10 @@
 - name: Upload extra files
   include_role:
     name: upload_extra_files
+
+#
+# Link uploads to /uploads if need be.
+#
+- name: upload the extra file info
+  shell: "ssh -oStrictHostKeyChecking=no {{ dyn_data.ssh_i_option }} {{ config_info.test_user }}@{{ dyn_data.test_hostname }} \"sudo ln -sf {{ dyn_data.kit_upload_directory }}/uploads /uploads\""
+
--- a/ansible_roles/roles/upload_extra/tasks/main.yml
+++ b/ansible_roles/roles/upload_extra/tasks/main.yml
@@ -21,16 +21,21 @@
     line: "status: success"
     create: yes
 
+- name: include dynamic info
+  include_vars:
+    file: "{{ working_dir }}/ansible_run_vars.yml"
+    name: dyn_data
+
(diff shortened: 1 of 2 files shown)


ansible_roles/roles/upload_extra/tasks/main.yml
```diff
@@ -27,3 +27,10 @@
 - name: Upload extra files
   include_role:
     name: upload_extra_files
+
+#
+# Link uploads to /uploads if need be.
+#
+- name: upload the extra file info
```
frival: I'm pretty sure we can't use /uploads for this as we have to support RHEL Image Mode which has a read-only root.  @kdvalin comments?

dvalinrh: /uploads has been expected since day 1.  If we need to move it, we can do so, it just has to be in a well known location.

ansible_roles/roles/upload_extra/tasks/main.yml
```diff
@@ -21,16 +21,21 @@
...
+- name: include dynamic info
+  include_vars:
+    file: "{{ working_dir }}/ansible_run_vars.yml"
+    name: dyn_data
+
+- name: make the uploads directory
+  shell: "ssh -oStrictHostKeyChecking=no {{ dyn_data.ssh_i_option }} {{ config_info.test_user }}@{{ dyn_data.test_hostname }} mkdir -p {{ dyn_data.kit_upload_directory }}"
+
+- name: Now link to /
+  shell: "ssh -oStrictHostKeyChecking=no {{ dyn_data.ssh_i_option }} {{ config_info.test_user }}@{{ dyn_data.test_hostname }} ln -s {{ dyn_data.kit_upload_directory }}/uploads /uploads"
```
kdvalin, line 33: This will fail on bootc hosts, discussion needed

kdvalin (changes requested):

dvalinrh:
Closing this, as this is not the proper fix for this problem.

PR #327, opened 2025-11-14
Remove the ssh perm file from the run directory once the test is

### LINKED ISSUE #326: Do not save user.pem_test in results area of zathras.
user.pem_test contains private key info, do not save it in the results area.

dvalinrh:
complete.

# Description
When the test is complete, it will remove the user perm file for ssh from the run area.

# Before/After Comparison
Before:  The ssh private key was left in the run area for Zathras.
After:  File is no longer in the run area of Zathras, preventing   accidental copy else weres.

# Documentation Check
No
# Clerical Stuff
This closes #326 

Relates to JIRA: RPOPC-703

### DIFF
--- a/bin/kick_off.sh
+++ b/bin/kick_off.sh
@@ -178,7 +178,7 @@ done
 remove_perm_file()
 {
 	if [[ $ssh_key_file != "" ]]; then
-		rm -rf config/user.pem_test
+		rm -f config/user.pem_test
 	fi
 }
 
--- a/bin/kick_off.sh
+++ b/bin/kick_off.sh
@@ -173,6 +173,16 @@ do
 done
 
 #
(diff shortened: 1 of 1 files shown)


bin/kick_off.sh
```diff
@@ -172,6 +172,16 @@ do
 	let "tests=$tests+1"
 done
 
+#
+# If we created the permission file, remove it when done.
+#
+remove_perm_file()
+{
+	if [[ $ssh_key_file != "" ]]; then
+		rm -rf config/user.pem_test
```
kdvalin: Drop the `-r` flag

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #330, opened 2025-11-17
Remove java

### LINKED ISSUE #329: Need to revisit designating java versions
Right now java versions are installed by either the user or directly by Zathras.  With the new packaging tool, where the wrapper installs the packages for us, we need to address how we do this and handle multiple versions of java.

dvalinrh:
# Description
Syncs up with the specjbb wrapper doing the java install

# Before/After Comparison
Before: Using the old method to install java, as well as the new method.
After:  The wrapper now installs the required java, via package_tool.

# Documentation Check

# Clerical Stuff
This closes #329 

Relates to JIRA: RPOPC-658


dvalinrh:
Closed, as additional changes creeped  in, redoing request.

PR #333, opened 2025-11-18
Remove java from zathras

### LINKED ISSUE #329: Need to revisit designating java versions

dvalinrh:
# Description
Removes Zathras doing the installation of java, that is now the realm of the wrappers.  As part of this, we update the test_defs specjbb entry to include --java_version

# Before/After Comparison
Before: Zathras installed java.
After: Zathras no longer installs java, instead relies on the wrappers to do it.

# Documentation Check
Yes, updated.

# Clerical Stuff
#
This closes #329

Relates to JIRA: RPOPC-658


kdvalin (approved):
LGTM

kdvalin:
The reference issue is a PR.  Please fix

PR #337, opened 2025-12-04
Add Container image for Zathras

### LINKED ISSUE #336: Create container image for Zathras
In some instances standing up a Zathras setup can be a major friction point (usecases like running Zathras in a CI/CD pipeline).

We should build a container image with Zathras set and ready to go.

kdvalin:
# Description
This PR adds a container build process for Zathras.  This will make it much easier to use Zathras in CI/CD enviornments or for people who do not want to deal with setting up their system with all the required dependencies Zathras requires.

This PR also separately allows for the `AWS_DEFAULT_REGION` env var to set the AWS enviornment if a AWS configuration is not found.  This is useful in CI/CD enviornments where credentials and configurations are primarily provided through enviornment variables.

# Before/After Comparison
## Before
The user had to install the dependencies required for Zathras through the install script, which could be problematic if they had a newer `yq` or `jq` version.  CI/CD enviornments need to stand up the dependencies manually.

CI/CD users needed to build an AWS config within the container.

## After
The user can simply run everything from the `quay.io/zathras/zathras` container.

CI/CD users can use `AWS_DEFAULT_REGION` instead of building an AWS config file.

# Documentation Check
Docs will be needed, will come soon

# Clerical Stuff
This closes #336 

Relates to JIRA: RPOPC-711

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.github
+docs
+cli_examples
+ci
--- a/.github/workflows/build_container.yaml
+++ b/.github/workflows/build_container.yaml
@@ -1,0 +1,41 @@
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  containerize:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Login to Quay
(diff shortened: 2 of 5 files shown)


dvalinrh (approved):
LGTM

PR #339, opened 2025-12-08
Fix it so uploads will only be created on writeable filesystems.

### LINKED ISSUE #338: uploads need to be created on a writeable filesystem, not just the one with the most free space
Need to use the filesystem that is writeable for uploads.

dvalinrh:
# Description
Uploads need to be created on a filesystem that is writeable.  Issue from bootc

# Before/After Comparison
Before:  If the filesystem that had the most free space was readonly, the upload creation/operation will fail.
After:  Zathras now filters out the read only filesystems for upload creation, so upload is not created there.

# Documentation Check

# Clerical Stuff
This closes #338
Relates to JIRA: RPOPC-741

Testing.
Ran to system that filesys with the most free space was read only, Zathras did not attempt to use it.
Verified that uploading of results still works properly.

### DIFF
--- a/tools_bin/writeable_filesys
+++ b/tools_bin/writeable_filesys
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+most_free_space=0
+usage()
+{
+	echo "Usage $1:"
+	echo "No option, shows the following for all writeable filesystems (not tmpfs)"
+	echo " <free_space> <mount point> <device>"
+	echo "  --most_free_space:  provide the mount point of the filesystem with the most free space"A
+	echo "  -h --usage: this usage message"
+        exit 0
+}
+
+NO_ARGUMENTS=(
+	"most_free_space"
+	"usage"
+)
+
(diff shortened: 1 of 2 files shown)


tools_bin/writeable_filesys
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+most_free_space=0
+usage()
+{
+	echo "Usage $1:"
+	echo "No option, shows the following for all writeable filesystems (not tmpfs)"
+	echo " <free_space> <mount point> <device>"
+	echo "  --most_free_space:  provide the mount point of the filesystem with the most free space"A
```
kdvalin: typo

kdvalin (changes requested):
Verify documentation changes are not needed

dvalinrh:
Does not require documentation update.  The location to upload to is not an option that the user can designate.

kdvalin (approved):
LGTM

PR #341, opened 2025-12-10
[Container] Add aarch64 image type

### LINKED ISSUE #340: [Container] Add multi-arch support
Sometimes it's desirable to run Zathras in a container on aarch64, we should support this.

kdvalin:
# Description
Adds an aarch64 container to be built along with the x86_64 container.

# Before/After Comparison
## Before
There is no aarch64 container build

## After
An aarch64 container is built along with x86_64 container under the same name.

# Documentation Check

Nope

# Clerical Stuff
Closes #340 

Relates to JIRA: RPOPC-752


grdumas (commented):
LGTM

grdumas (approved):
LGTM, part two

PR #344, opened 2025-12-16
Add in tag default file.

### LINKED ISSUE #342: add a tag.config file that zathras will look for in the users home directory.
Add ability to pull the tag config file information from the users home directory.  Contents will override what is in config/tags*.

dvalinrh:
# Description
Assigning tags will now check .config/zathras/cloud_tags for tags to assign.

# Before/After Comparison
Before: Required the user to always update tags.config to set the tags
After: The user now is able to set the values once in the .config/zathras/cloud_tags file, and the values will be used
            everytime

# Documentation Check
Yes, updates made.

# Clerical Stuff
This closes #342 

Relates to JIRA: RPOPC-753

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -1533,7 +1533,7 @@ generate_tags()
   		echo "  Jirald: ${1}" >> tags_defaults
 	fi
 	#
-	# If ~/.config/zathras/cloud_tags exsists, replace the various tags.
+	# If ~/.config/zathras/cloud_tags exists, replace the various tags.
 	# 
 
 	rm -f tag_temp
--- a/bin/burden
+++ b/bin/burden
@@ -1533,6 +1533,28 @@ generate_tags()
   		echo "  Jirald: ${1}" >> tags_defaults
 	fi
 	#
+	# If ~/.config/zathras/cloud_tags exsists, replace the various tags.
+	# 
+
+	rm -f tag_temp
+	if [[ -f "${HOME}/.config/zathras/cloud_tags" ]]; then
+		while IFS= read -r line
+		do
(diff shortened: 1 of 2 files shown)


bin/burden
```diff
@@ -1533,6 +1533,28 @@ generate_tags()
...
+	# 
+
+	rm -f tag_temp
+	if [[ -f "${HOME}/.config/zathras/cloud_tags" ]]; then
+		while IFS= read -r line
+		do
+			name=`echo $line | cut -d':'  -f 1`
+			while IFS= read -r tag_value
+			do
+				field=`echo $tag_value | cut -d':' -f 1`
```
kdvalin:
```suggestion
				field=`echo "$tag_value" | cut -d':' -f 1`
```

bin/burden
```diff
@@ -1533,6 +1533,28 @@ generate_tags()
...
+				echo $field | grep -q $name
```
kdvalin:
```suggestion
				echo "$field" | grep -q "$name"
```

bin/burden
```diff
@@ -1533,6 +1533,28 @@ generate_tags()
   		echo "  Jirald: ${1}" >> tags_defaults
 	fi
 	#
+	# If ~/.config/zathras/cloud_tags exsists, replace the various tags.
```
kdvalin:
```suggestion
	# If ~/.config/zathras/cloud_tags exists, replace the various tags.
```

kdvalin (changes requested):

kdvalin (approved):
LGTM

PR #353, opened 2026-02-24
Add kit directory validation during installation

### LINKED ISSUE #323: zathras should warn if /home/uploads directory is missing
If /home/uploads doesn’t exist during installation, wrappers depending on uploaded kits may fail silently. Add a non-blocking warning to improve visibility.

Proposed Fix:
- Check for /home/uploads in the install script.
- If missing, print:
Warning: /home/uploads directory not found. Some wrappers may fail until it is created.

sayalibhavsar:
# Description
This PR adds inline validation to install.sh that checks for missing kit files/directories specified in config templates and displays clear warnings during installation.

# Before/After Comparison
Before: Installation completed silently even if required kits were missing, causing wrappers to fail later without clear indication why.
After: Installation displays a formatted warning listing all missing kits with their paths and guidance on how to resolve the issue.

# Documentation Check
No documentation updates needed as this is a non-blocking warning feature that enhances existing installation behavior without changing user workflows.

# Clerical Stuff
This closes #323
Relates to JIRA: RPOPC-700

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -209,6 +209,76 @@ write_installation_record() {
 # Write installation record
 write_installation_record
 
+# Validate kit directories specified in templates
+echo ""
+echo "Validating kit directories..."
+
+missing_kits=()
+config_dir=$(readlink -f "config")
+
+for template_file in config/*_template.yml; do
+    # Skip if no files match the pattern
+    [ -e "$template_file" ] || continue
+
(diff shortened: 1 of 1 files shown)


grdumas (approved):
LGTM

dvalinrh (commented):
lgtm

PR #365, opened 2026-03-10
Shakedown fixes

### LINKED ISSUE #364: update to zathras for issues shook out from shakedown run
Issues uncovered

Issue with writeable_filesys on ubuntu
updating test versions
issue with git not being installed with OS.
issue with SUSE installing Suse specific package.

dvalinrh:
# Description
During the shakedown run a couple issues turned up
Issue with writeable_filesys on ubuntu
Older test versions
issue with git not part of the OS install
issue with SUSE installing Suse specific package.

This pr addresses those issues.

# Before/After Comparison
Before: Various failures across SUSE, Ubutnu and Amazon 2023
After: All wrappers tested, ran properly on all OS flavors

# Documentation Check
No

# Clerical Stuff
This closes #364 

Relates to JIRA: RPOPC-861

Testing done
Issue with writeable_filesys on ubuntu
    Ubuntu was pulling the wrong fileystem due to what devices where being shown by df.  Fix was to go to findmnt
    Ran speccpu on all flavors of OS, proper disk was chosen.

Older test_versions
    Updated to the latest versions and verified they worked.

git not part of the OS install
    If git is not part of the OS install, we can not install test_tools-wrappers.  Solution is to install git just before we
    we execute the test wrapper.
    Verified fix ran on all OS flavors, and OS having the issue, no longer experienced the issue.

 issue with SUSE installing Suse specific package.  
    Certain versions of SUSE does not have sle-module-legacy.  Fix is to no longer terminate, but simply warn about
    the failure.


dvalinrh:
For qudo flagging findmnt
We have added a grep filter to filter out nfs etc filesystems.

bin/ten_of_us.yml
```diff
@@ -366,19 +366,18 @@
...
+ 
 - hosts: local
   vars_files: ansible_vars.yml
   gather_facts: no
   tasks:
-  - name: Terminate on SLES legacy module failure
+  - name: Warning of sles_legacy
     include_role:
-      name: terminate_on_error
+      name: warn_on_error
```
frival: Making this a warning instead of an error means we're going to fall much, much further down into the automation before we fail in the affected benchmark wrapper.  That violates the principal of failing as early as possible.  I don't think we've solved anything here, we've just made the user's life more difficult.  Further, the justification that "certain versions of SUSE does not have sle-module-legacy" does not match Suse's own documentation so if the legacy module is not found something else is wrong with the image and declaring failure is probably the right thing to do.

tools_bin/writeable_filesys
```diff
@@ -53,14 +53,19 @@ worker_fsys=`mktemp /tmp/zathras_fs.XXXXX`
 out_file=`mktemp /tmp/zathras_fs.XXXXX`
 mount | grep ' (rw,' > $worker_mnt
 
-df --local --output=avail,target,source | grep -ve 'mounted\|tmpfs\|boot' | grep -v "Mounted on" > $worker_fsys
+#
+# Filter out remote filesystems.
+#
+findmnt -D -t nosquashfs,notmpfs,nodevtmpfs,noproc,nosysfs -b | grep -vE 'nfs|cifs|smb|sshfs|fuse.sshfs| lustre' > $worker_fsys
```
frival: Do you mean to have the space before lustre?

frival (changes requested):
Apparent typo, and we need to talk more about this SLES change.

dvalinrh:
Suse returned to prior behavior of terminating if can not install legacy package.
git now is checked to see if it is present.  If not we attempt to install and bail if we can not install the git package.  This piece is a temporary fix, issue https://github.com/redhat-performance/zathras/issues/366

frival (approved):
LGTM

PR #368, opened 2026-03-13
Adding support for rerunning certain failed tests.

### LINKED ISSUE #367: Update rerun feature
Updating rerun feature to rerun only tests that have returned an error code between 1 and 100.

dvalinrh:
# Description
Improve the rerun feature of failed tests.  Earlier work was a primitive rerun of the test, did we produce an error, yes rerun the test.  We now have a range on errors to rerun (1-99), as well as handle multiple failures with in a list of tests to run from

# Before/After Comparison
Before: Primitive rerun ability of tests.
After: More selective of what tests we will rerun.

# Documentation Check
None

# Clerical Stuff
This closes #367 

Relates to JIRA: RPOPC-866

Testing
Ran coremark and streams with an additional option of --induce_error

Test run with the option of induce_error
streams set to 1
coremark set to 1
  Results:  Test ran as expected, ran each test twice.

streams set to 101
coremark set to 1
   Results:  Only coremark was executed twice

streams set to 1
coremark set to 0
   Results only streams ran twice.

No induce_error setting
  Tests only ran once.

debug file:
[report_out.txt](https://github.com/user-attachments/files/25966231/report_out.txt)


dvalinrh:
For this line
Multi-line failure parse 🐞 Bug ⛯ Reliability
handle_failure_reruns uses awk over the entire test_failure_* file and assumes single-line
output for fcommand and rtc; if the failure file contains multiple lines, rtc becomes
non-numeric/multi-valued and the -gt 99 check can error or skip reruns. This situation is likely
when multiple variant tests share the same exec_dir/test_failure and write multiple distinct lines
into it.

This is controlled, and there will only be one line ever in the test_failure* file.  If there isn't then we have a bug elsewheres.

kdvalin:
Closing this out due to recent history rewrite (private SSH key accidentally merged)

PR #374, opened 2026-03-20
Remove the install status files before performing the requested

### LINKED ISSUE #373: remove the install status file before performing the operation.
Issue appeared where Zathras was failing because of prior failure recorded in tar status.  Remove the tar_status file before doing anything else in the role install_tools

dvalinrh:
# Description
What does this PR do?  Give a short summary
Remove the install status files before performing the requested operation

# Before/After Comparison
Before:  If a prior run of zathras has resulted in an install failure, we will keep picking up on it in subsequent runs.
After:  The install status in the status file will be from the current run.

# Documentation Check
No

# Clerical Stuff
This closes #373 

Relates to JIRA: RPOPC-891

Testing Done
Reproduced the issue, and got a bogus failure
After change
Ran the changes on account that was seeing the issue, no longer present (bare metal system)
Ran on a cloud system, everything is fine.

### DIFF
--- a/bin/ten_of_us.yml
+++ b/bin/ten_of_us.yml
@@ -228,7 +228,7 @@
 
   gather_facts: no
   tasks:
-  - name: Remove the tar_status file if it exists
+  - name: Remove the cr_status file if it exists
     ansible.builtin.file:
       path: /tmp/cr_status
       state: absent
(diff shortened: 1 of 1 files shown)


malucius-rh (approved):
LGTM
Caused failure, made repairs, second run now able to complete successfully

PR #378, opened 2026-03-24
Update the test versions

### LINKED ISSUE #377: update test versions to latest.
Update the versions of the test in config/* to use the latest versions.

dvalinrh:
# Description
Brings all the tests defined in config to be the latest version

# Before/After Comparison
Before: Ran with the older version of the wrapper
After: Runs now with the latest version of the wrapper.

# Documentation Check
No

# Clerical Stuff
This closes #377 

Relates to JIRA: RPOPC-897

Testing
All the new versions of the wrapper has been tested when the wrapper was pushed.
When test is executed, we find the proper version of all the wrappers (bad version of any wrapper, gets flagged).


frival (approved):
LGTM

PR #382, opened 2026-04-14
Fix burden not honoring waiting for pids all the time

### LINKED ISSUE #380: zathras is not working 100% correctly with waiting for children to finish.
Couple cases uncovered where we were not waiting for all the children has been uncovered.  This corrects that issue.

dvalinrh:
# Description
Fixes things so we wait for all the children.

# Before/After Comparison
Before: Cases were seen where we are not honoring waiting for the children.
After: Now honors waiting for the children.

# Documentation Check
None

# Clerical Stuff
This closes #380 

Relates to JIRA: RPOPC-927

Tested:
Verified that it works properly with the environment seen in.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -288,8 +288,8 @@ gl_tf_sys_list=`mktemp /tmp/zathras_cl_systems.XXXXX`
 
 wait_for_procs()
 {
-	pids=$1
-	for pid in "${pids[*]}"; do
+	pids=("$@")
+	for pid in "${pids[@]}"; do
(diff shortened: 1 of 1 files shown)


bin/burden
```diff
@@ -320,6 +320,7 @@ cleanup_and_exit()
 	rtc=$2
 	rtr_string=""
 
+
```
frival, line 323: Spurious blank line

frival (approved):
LGTM, the blank line isn't worth blocking for.

PR #392, opened 2026-05-26
Fix aws output

### LINKED ISSUE #391: Set relevant output types for cloud
AWS will break on anything but text output mode.  
One solution is to set the AWS_DEFAULT_OUTPUT to be text

Do we do the same things for the clouds?

dvalinrh:
# Description
1) Sets AWS_DEFAULT_OUTPUT to be text, so the output from the aws cli commands are as expected.
2) In cleanup_and_exit check to see if the instance exists before attempting to use terraform to terminate it.  Elienating a bogus error message.

# Before/After Comparison
Before:  If json was chose for the aws cli output, zathras will fail because the output is not in the proper format.  Also a bogus error message is put to the screen concerning attempting to delete the non-existent image.
After: The aws cli output is now in the expected format, and Zathras runs properly.  The bogus error message is no longer given if we exit out before the instance is actually created.

# Documentation Check
No

# Clerical Stuff
This closes #391 

JIRA: RPOPC-1236

Testing:
Verified the proper format is used regardless of the users default output.
Verified that the bogus error message no longer appears.

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -50,6 +50,8 @@
 cli=${0}
 arguments=("$@")
 version="3.2"
+AWS_DEFAULT_OUTPUT="text"
+export AWS_DEFAULT_OUTPUT
 #
 # This bash script generates the test parameters to be used by ansible. End results
 # is an ansible data file called ansible_vars_main.yml containing the following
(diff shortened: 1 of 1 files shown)


kdvalin (approved):
LGTM

PR #393, opened 2026-05-26
Claude assisted: Add private subscription support for Azure.

dvalinrh:
# Description
Adds support to azure for booting private subscriptions

# Before/After Comparison
Before:  Unable to boot a private generated image.
After: Can now boot a private generated image.

# Documentation Check
Yes.  Updates where made.

# Clerical Stuff
This closes #390 

Relates to JIRA: RPOPC-1215

Testing
Verified can boot a private image, and runs occur properly.
Verified still can boot a public image.


grdumas:
@dvalinrh Could you include the logs showing the before and after behavior?

dvalinrh:
Real simple.  The before behavior we would not see or be able to boot the private ami.  After behavior we can.

dvalinrh:
Already in place.

PR #395, opened 2026-05-26
fix: Allow for arm64 and aarch64 to be used interchangably

### LINKED ISSUE #394: Aws AMI detection does not match aarch64 with arm64
See title, AMIs need to have `arm64` in their name even though `aarch64` is also valid.

We should treat these 2 arch names as the same, since they are identical.

kdvalin:
# Description
Allows AMI names to use arm64 or aarch64 for ARM runs.

# Before/After Comparison
## Before
AMIs had to have `arm64` in their name for ARM machines.

## After
AMIs can have `arm64` or `aarch64` in their name for ARM machines.

# Documentation Check

No doc changes needed.

# Clerical Stuff
This closes #394 

Relates to JIRA: RPOPC-1238

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -1459,6 +1459,12 @@ aws_specific_os_version()
 					break
 				fi
 			fi
+			# Clean up ambiguation between arm64 and aarch64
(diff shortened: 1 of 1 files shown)


dvalinrh (approved):
LGTM

PR #398, opened 2026-06-02
Handle cap of multiple systems properly.

### LINKED ISSUE #381: max bumber of procs has to apply for all system stanzas
If sys barrier is not in place, zathras will move onto the next system stanza.  This is by design, however the max_system counter is set up on a system stanza, not for everything.  We can inadvertently ending up with many more children then the max systems says to allow.

This change will require locking on an exported variable, so it is not straight forward as one would hope.

dvalinrh:
# Description
Right now each test entry in the scenario file has its own accounting for number of systems active.  This addresses the issue where multiple test entry in the scenario file, not bounded by a SYS_BARRIER may result in running more then the max number of systems you want executed.  This is accomplished by using the Zathras built in locking operations.

# Before/After Comparison
Scenario snippet
systems:
  system1:
    tests: "auto_hpl"
    host_config: "m5a.24xlarge,m6i.24xlarge,m6i.4xlarge"
  system2:
    tests: "coremark"
    host_config: "m6i.xlarge,m7a.16xlarge,m7a.24xlarge"

Before: 
   Running the scenario above would result in 6 vms being created at once, not the default cap of 3.
After:
   We will run a max of the default cap of 3

# Documentation Check
No

# Clerical Stuff
This closes #381 

Relates to JIRA: RPOPC-928

Testing:
Verified that the maximum number of systems is honored regardless how the scenario file is built.
Verified we get results for what we asked for.


bin/utils/gl_lock_ops
```diff
@@ -0,0 +1,34 @@
...
+gl_unlock_exclusive()
+{
+	gl_lock=${1}
+	local -n have_locked=$2
+	if [[ $have_locked -eq 1 ]]; then
+		if [[ $gl_debug_lock -eq 1 ]]; then
+			echo ${FUNCNAME[2]} ${BASH_LINENO[1]}
+		fi
+		rmdir $gl_lock
+		have_locked=0
```
frival, line 32: Stupid question, but what good does this do since have_locked is declared local and the next thing we do is return from the function?

bin/burden
```diff
@@ -207,6 +205,9 @@ gl_test_override_options=""
 gl_warning_string=""
 gl_max_systems=3
 gl_max_systems_set=0
+if [ ! -v $gl_total_systems ]; then
```
frival: This should be "if [ ! -v gl_total_systems ]; then" - -v is supposed to check if the variable is set, not the value of the variable.

bin/utils/gl_lock_ops
frival: I missed the fact that the "local -n" at line 4 makes this a nameref so it dereferences back to the caller's variable scope.  We're good here.

dvalinrh:
Timestamps on results file, max systems set to 3

drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:09 m6a.xlarge_1
drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:10 m5.xlarge_1
drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:12 m5a.xlarge_1
drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:18 m7a.xlarge_1
drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:23 m6i.xlarge_1
drwxr-xr-x. 3 dvalin dvalin 4096 Jun 24 12:24 m7i.xlarge_1

Note jump from 12 to 18, in state of waiting for systems to finish up.

frival (approved):
LGTM

PR #418, opened 2026-06-30
Re-enable zone selection within AWS

### LINKED ISSUE #416: Zone selection is broken
When trying to use the optional zone selection (to avoid things like unavailable instance types in certain availability zones), the following error occurs

```
Tue Jun 23 10:49:05 AM EDT 2026 n42-h30-000-r650.rdu3.labs.perfscale.redhat.com, Error: following fields in scenario has no entry in config/zathras_scenario_vars_def:
    host_config: m6g.4xlarge[region=us-east-1&zone=a]
```

kdvalin:
# Description
Availability zone selection was previously broken due to several factors.  This PR re-enables the ability to select availability zones.

Short and sweet: 
1. Validation looks for [ and errors when that's found, which is what's
   used for zone/region selection.
2. Provided zone would be overridden.
3. Several locations where the instance type does not filter out the
   `[...]` used to select the region..

# Before/After Comparison
## Before
Using documented `<instance type>[region=<region>&zone=<zone>]` would result in 
```
Error: following fields in scenario has no entry in config/zathras_scenario_vars_def:
    host_config: m6g.4xlarge[region=us-east-1&zone=a]
```

## After
Zone/region selection works as intended.

# Documentation Check
No documentation changes are needed, this is fixing a documented feature.

# Clerical Stuff
This closes #416
Relates to JIRA: RPOPC-1331

[az_specified.log](https://github.com/user-attachments/files/29511315/az_specified.log)
[no_specified.log](https://github.com/user-attachments/files/29511316/no_specified.log)

### DIFF
--- a/bin/burden
+++ b/bin/burden
@@ -1499,7 +1499,7 @@ azure_specific_os_version()
 
 aws_specific_os_version()
 {
-	vmtype=`echo $1 | cut -d':' -f 1`
+	vmtype=`echo $1 | cut -d':' -f 1 | cut -d'[' -f1`
 	ami_found=0
 
 	#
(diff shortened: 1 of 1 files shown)


dvalinrh (commented):
Please provide run logs for AWS.

dvalinrh (approved):
LGTM

PR #441, opened 2026-07-29
fix(azure): Move disks.tf when there's actually disks to work with

### LINKED ISSUE #413: Azure disk handling task fires even when no disks are configured
## Problem

The "Handle disks" block in `ansible_roles/roles/azure_create/tasks/main.yml` runs even when the scenario doesn't include extra disks, breaking scenarios that don't need disk configuration.

## Impact

Azure scenarios without extra disks may fail or behave unexpectedly during VM creation.

## Location

`ansible_roles/roles/azure_create/tasks/main.yml` — "Handle disks" block.

## Suggested fix

Add a guard condition:

```yaml
when: config_info.cloud_disks != "[0:na:na:0]" and config_info.cloud_numb_networks == 0
```

kdvalin:
# Description
Adds a simple check to see if the cloud_disks is "none"

# Before/After Comparison
## Before
On Azure, when `cloud_disks` is "none" the handle disks block in `ansible_roles/roles/azure_create/tasks/main.yml` would run, copying the `disk.tf` file when it's not needed.

## After
The entire handle disk block is skipped when there are no disks to deal with

# Documentation Check

No, internal bugfix

# Clerical Stuff
Closes #413 

Relates to JIRA: RPOPC-1330

### DIFF
--- a/ansible_roles/roles/azure_create/tasks/main.yml
+++ b/ansible_roles/roles/azure_create/tasks/main.yml
@@ -85,7 +85,7 @@
     copy:
       src: "tf/disks.tf"
       dest: "{{ working_dir }}/tf/disks.tf"
-  when: config_info.cloud_disks != "[0:na:na:0]"
+  when: config_info.cloud_disks != "[0:na:na:0]" and config_info.cloud_disks != "none"
 
 - name: include variables etc.
   include_vars:


kdvalin:
@coderabbitai review

sayalibhavsar (approved):
LGTM

Issue #428, opened 2026-07-14
[IBM] instances not cleaned up after mid-run failure


sayalibhavsar:
If terraform apply gets partway through (creates the subnet and security group) then dies on a name collision or quota error, nothing cleans upafter it. terminate_cloud only runs on success, so the partial resources just sit there in IBM Cloud. 

The leftover names cause collisions on the next run.

Fix: run terraform destroy on the failure path too. If apply fails, the role should still try to tear down whatever it created.

sayalibhavsar:
Reproducer 

run1 https://gist.github.com/sayalibhavsar/677c2efef6aba5d663ed9d7d790baf7c 
https://drive.google.com/file/d/1jFzDfZQ72uf7b8lpAfB-qVXsMDZc9T6V/view?usp=sharing
run2 https://gist.github.com/sayalibhavsar/8692165baf03e07b150468afeea1ffb2

./bin/utils/terraform_ops --terminate_list all cleaned it up https://gist.github.com/sayalibhavsar/01fa18e35e18654df011d0f8a7ecc33e

kdvalin:
This is looking a lot like #387, Zathras goes off the happy path and as a result, cloud resources are left running.

kdvalin:
@sayalibhavsar has this issue been fully resolved?

sayalibhavsar:
Yes this has been solved with a combination of adding random suffix to all resources and midrun run cleanup sigterm fix

kdvalin:
Closing this out then

Issue #387, opened 2026-05-07
Azure: orphaned resource groups when CI/CD pipeline times out


sayalibhavsar:
## Problem

When running burden in a CI/CD pipeline (e.g., GitLab CI) and the job hits the pipeline timeout limit, the job is killed before burden can run terraform destroy. This leaves orphaned Azure resource groups containing VMs, disks, NICs, and public IPs, incurring ongoing costs until manually deleted.

## Expected behavior

Azure resources should be cleaned up even when the pipeline job is terminated by a timeout.

## Steps to reproduce

1. Run burden in a GitLab CI pipeline with `terminate_cloud: 1` in the scenario
2. Set a pipeline timeout shorter than the benchmark runtime
3. The job gets killed at timeout
4. The Azure resource group (e.g., `root-RHEL-10.2-...-D4ps-v5-00`) remains with all resources still running

## Impact

- Orphaned VMs and Premium SSD disks continue to accrue Azure charges
- Manual cleanup required via `az group delete --name <resource-group> --yes`
- No warning or notification that resources were left behind

## Suggested fix

Add an `after_script` or trap-based cleanup in the pipeline templates that runs even on timeout/failure. For example:

```yaml
after_script:
  - az login --service-principal -u $AZURE_CLIENT_ID -p "$AZURE_CLIENT_SECRET" --tenant $AZURE_TENANT_ID
  - az group list --query "[?starts_with(name, 'root-')].name" -o tsv | xargs -I{} az group delete --name {} --yes --no-wait
```

Or within burden itself, register a signal handler (SIGTERM/SIGKILL) that triggers terraform destroy before exit.

## Environment

- Zathras container: `quay.io/zathras/zathras:latest`
- Cloud provider: Azure
- CI/CD: GitLab CI
- Scenario: `system_type: azure` with `terminate_cloud: 1`

kdvalin:
Would using `bin/utils/terraform_ops --terminate_list all` fix this problem?  We should keep this open to fix SIGTERM signal handling, but that would be a more elegant workaround if it functions.

kdvalin:
If that workaround does work, then I would bump this to a later version.

sayalibhavsar:
Yeah, `terraform_ops --terminate_list all` works for manual cleanup most of the time.
I was thinking of az commands in case terraform state is corrupted or missing and many times the orphaned resources survive. But it relies on naming convention and might accidentally cleanup unrelated resource groups.

[...]

PR #444, opened 2026-08-11
fix: clean up cloud resources on midrun failure

sayalibhavsar:
Enable uperf on IBM Cloud and clean up cloud resources on midrun terraform failure.

# Before/After Comparison

Before: Uperf fails on IBM Cloud — server/client IPs aren't recorded and VMs can't SSH to each other. Terraform partial failures and pipeline timeouts leave orphaned VMs.

After: IPs recorded, SSH keys exchanged between VMs, partial terraform failures trigger cleanup, and burden traps SIGTERM for proper teardown.

# Documentation Check
No.

# Clerical Stuff
Relates to JIRA: RPOPC-<TICKET_NUMBER>
Solves: #387 #428

### DIFF
--- a/ansible_roles/roles/tf_create/tasks/tf_apply.yml
+++ b/ansible_roles/roles/tf_create/tasks/tf_apply.yml
@@ -32,6 +32,14 @@
     line: "rtc: {{ '0' if tf_success else '1' }}"
     backrefs: yes
 
+- name: Clean up partial resources on failure
+  when: not tf_success
+  include_role:
+    name: tf_delete
+  vars:
+    tf_dir: "tf"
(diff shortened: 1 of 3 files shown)


kdvalin:
Some comments outside the scope of the diffs, we are checking for a running state for the VMs before allowing terraform_ops to kill it.

@dvalinrh / @frival Is there a reason why this check exists?  It is limited to the scope of the running burden instance, shouldn't be delete all instance types, regardless of their running state in the case of a signal interrupting burden?

kdvalin (approved):
LGTM, other points can be taken care of later.

